PR #17, opened 2019-10-31
Feature/cvsb 9465

irina005:


.reports/test-report.xml
```diff
@@ -0,0 +1,72 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
arupdvsa: .reports folder need to be added to .gitignore

src/app/components/technical-record/technical-record.component.scss
```diff
@@ -150,7 +151,7 @@ dd {
 
 .title {
   color: $govuk-text-colour;
-  font-family: $font-stack;
+  font-family: "GDS Transport",Arial,sans-serif;
```
arupdvsa: as per spec font family should be GDSTransportWebsite
image
Any other aspect need to be handled through specificity

PR #29, opened 2019-11-11
Feature/CVSB-9654&CVSB-9655&CVSB-9650&9465

irina005:
Solved bugs - CVSB-9655, CVSB-9650, CVSB-9465

### DIFF
--- a/src/app/components/technical-record/technical-record.component.ts
+++ b/src/app/components/technical-record/technical-record.component.ts
@@ -53,4 +53,8 @@ export class TechnicalRecordComponent implements OnInit {
     return (typeof str==='string' || str==null) ? !str||!str.trim():false;
   }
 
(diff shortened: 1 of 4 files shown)


src/app/components/technical-record/technical-record.component.ts
```diff
@@ -53,4 +53,8 @@ export class TechnicalRecordComponent implements OnInit {
     return (typeof str==='string' || str==null) ? !str||!str.trim():false;
   }
 
+  hasProperty(obj, prop){
```
arupdvsa: why we need to implement `hasProperty`
Can we not instead use like below

```
<div *ngIf="selectedElement.hasOwnProperty('prop');then 
showMyDiv">...</div>

<ng-template #showMyDiv> 
my div
</ng-template> 
```

PR #55, opened 2019-12-14
Work on ngrx adrForm and tech record mapping for updating with dynamo data

irina005:
added new action for update form fields, effect for it, then refactored vehicleTechRec effects, refactored a pipe & added mapping function in tech rec component for adrForm & dynamo data mapping


src/app/pipes/FilterRecordPipe.ts
```diff
@@ -1,18 +1,14 @@
...
-      records = techRecordList.find((record: any) => record.statusCode === 'provisional');
-      if (records !== undefined) return records;
+  constructor(private techRecordService: TechnicalRecordService) {
+  }
 
-      records = techRecordList.filter((record: any) => record.statusCode === 'archived');
-      records.sort((a,b)=> new Date (a.createdAt).getTime() - new Date (b.createdAt).getTime() );
-      if (records !== undefined) return records[0];
-    }
+  transform(techRecordList: any): any {
```
arupdvsa: should avoid using service inside a pipe .. pipes are used in template. Therefore operations should be slim and invariantly pure else will side-effect into slow performance

src/app/store/effects/VehicleTechRecordModel.effects.ts
```diff
@@ -1,39 +1,49 @@
...
+    map(techRecordResponse => {
+      return new GetVehicleTechRecordModelHavingStatusAllSuccess(techRecordResponse)
+    }),
+    catchError((error) => {
+      return of(new GetVehicleTechRecordModelHavingStatusAllFailure(error));
+    })
+  );
+
+  @Effect({dispatch : false})
+  getTechnicalRecordsStatusSuccess$ = this._actions$.pipe(
```
arupdvsa: not sure what you are trying to do here, but nonetheless will not make sense with decorator `dispatch false` when in implementation you are dispatching out

src/app/technical-record-search/technical-record.service.ts
```diff
@@ -21,10 +22,21 @@ export class TechnicalRecordService {
 
   getTechnicalRecordsAllStatuses(searchIdentifier: string): Observable<any> {
     const headers = new HttpHeaders().set('Content-Type', 'application/json; charset=utf-8');
-    return this.httpClient.get<any[]>(this.routes.techRecordsAllStatuses(searchIdentifier), { headers });
+    return this.httpClient.get<any[]>(this.routes.techRecordsAllStatuses(searchIdentifier), {headers});
   }
 
-  uploadDocuments() {
+  getActiveTechRecord(techRecordList: any) {
```
arupdvsa: this code is merely doing transformation ... not an API service call

src/assets/config/config.dev.json
```diff
@@ -3,29 +3,27 @@
...
     "cacheLocation": "localStorage",
     "endpoints": {
       "api": "xxx-bae6-4760-b434-xxx"
     }
   },
   "apiServer": {
-    "APITechnicalRecordServerUri": "http://localhost:3005",
-    "APITestResultServerUri": "http://localhost:3006",
-    "APIDocumentsServerUri": "https://api.nonprod.cvs.dvsacloud.uk/BRANCH"
+    "APITechnicalRecordServerUri": "",
```
arupdvsa: this should be replaced with placeholders

PR #56, opened 2019-12-17
fixing design for tank documents

irina005:


src/app/technical-record/technical-record.component.ts
```diff
@@ -85,7 +94,7 @@ export class TechnicalRecordComponent implements OnInit {
   }[]>;
   tc3Inspections$: Observable<any[]>;
 
-  constructor(private _store: Store<IAppState>, public matDialog: MatDialog) {
+  constructor(private _store: Store<IAppState>, public matDialog: MatDialog, public router: Router) {
```
arupdvsa: we need to use ngrx-router state to tap router actions in effects , that way we can retain the SOLID design's first paradigm , which is separation of concern.

component is related with UI (View) rendering only , while we use effects to respond to changes due to user actions

arupdvsa: please can you remove angular router injection in the component

PR #81, opened 2020-02-06
feature/CVSB-11769 - Technical record refactoring

irina005:
https://jira.dvsacloud.uk/browse/CVSB-11769


src/app/adr-details-view/adr-details-view.component.ts
```diff
@@ -0,0 +1,53 @@
...
+  permittedDangerousGoodsFe$: Observable<string[]>;
+  guidanceNotesFe$: Observable<string[]>;
+
+  constructor(private _store: Store<IAppState>, public techRecHelpers: TechRecordHelpersService) {
+    this.vehicletypes$ = this._store
+      .pipe(select(s => s.vehicleTechRecordModel.vehicleTechRecordModel.metadata.adrDetails.vehicleDetails.typeFe));
+    this.permittedDangerousGoodsFe$ = this._store
+      .pipe(select(s => s.vehicleTechRecordModel.vehicleTechRecordModel.metadata.adrDetails.permittedDangerousGoodsFe));
+    this.guidanceNotesFe$ = this._store
+      .pipe(select( s => s.vehicleTechRecordModel.vehicleTechRecordModel.metadata.adrDetails.additionalNotes.guidanceNotesFe));
```
rspasici: move those into ngOnInit method [https://ultimatecourses.com/blog/angular-constructor-ngoninit-lifecycle-hook](Constructor vs ngOnInit)

src/app/adr-details-view/adr-details-view.component.ts
```diff
@@ -0,0 +1,53 @@
...
+import {DownloadDocumentFileAction} from '@app/adr-details-form/store/adrDetails.actions';
+import {select, Store} from '@ngrx/store';
+import {IAppState} from '@app/adr-details-form/store/adrDetailsForm.state';
+import {Observable} from 'rxjs';
+
+@Component({
+  selector: 'vtm-adr-details-view',
+  templateUrl: './adr-details-view.component.html',
+  styleUrls: ['./adr-details-view.component.scss'],
+  encapsulation: ViewEncapsulation.Emulated,
```
rspasici: i don't think we need those. emulated is the default option
https://angular.io/api/core/ViewEncapsulation

src/app/body/body.component.html
```diff
@@ -0,0 +1,47 @@
+<dl class="technical-record-sub-details govuk-summary-list ">
```
rspasici: This file should be unde technical record folder, not the root app folder

src/app/notes/notes.component.scss
```diff
@@ -0,0 +1 @@
+@import "src/app/app.component.scss";
```
rspasici: this is not needed here (nor into other components) if there style are in app.component. if it doesn't work that way just import in into styleUrls in the `component.ts` file, but as i said, `@imports` into component styles are only for importing styles from 3rd party libraries, not our own styles

src/app/technical-record-search/technical-record-search.component.scss
```diff
@@ -1,4 +1,5 @@
 @import "node_modules/govuk-frontend/govuk/all";
```
rspasici: also i do not think you need this if you just `@import` from the root style of the app `app.component.scss`

src/app/test-history/test-history.component.html
```diff
@@ -0,0 +1,42 @@
...
+    <tbody class="govuk-table__body">
+    <tr>
+      <td>-</td>
+      <td>-</td>
+      <td>-</td>
+      <td>-</td>
+      <td>-</td>
+    </tr>
+    </tbody>
+  </ng-template>
```
rspasici: why do we need this ?

src/app/vehicle-summary/vehicle-summary.component.ts
```diff
@@ -0,0 +1,30 @@
...
+export class VehicleSummaryComponent implements OnInit {
+
+  @Input() activeRecord: any;
+
+  constructor(public techRecHelpers: TechRecordHelpersService) { }
+
+  ngOnInit() {
+  }
+
+  public axlesHasNoParkingBrakeMrk(axles) {
```
rspasici: add `:boolean` at the end. also why is this public ?

src/app/adr-details-view/adr-details-view.component.ts
irina005: This is the way Angular generated my component

src/app/test-history/test-history.component.html
irina005: This template is displayed when there are no results in testResultJson

src/app/vehicle-summary/vehicle-summary.component.ts
irina005: Done

src/app/adr-details-view/adr-details-view.component.ts
rspasici: you can choose a path when you create angular component
https://stackoverflow.com/questions/42144953/how-can-i-change-the-path-of-component-generation-to-current-path-in-angular-cli

src/app/test-history/test-history.component.html
m-ujaffer: The default template should not be a hardcoded row. It should be mapped to the actual testResult object and the initialState of the object would be using the pipe which we can add for the "-".

PR #86, opened 2020-02-14
feature/CVSB-10139 - As the DVSA, we want to be able to search for a vehicle before creating

irina005:
https://jira.dvsacloud.uk/browse/CVSB-10139


src/app/technical-record-create/technical-record-create.component.ts
```diff
@@ -7,10 +14,37 @@ import { Component, OnInit, ChangeDetectionStrategy } from '@angular/core';
...
+      vrmControl.setValidators([Validators.minLength(1), Validators.maxLength(8)]);
+    }
+    vrmControl.updateValueAndValidity();
+  }
+
+  onSubmit() {
+    this._store.dispatch(new SetVehicleTechRecordModelVinOnCreate({
+      vin: this.createTechRecordForm.get('vin').value,
+      vrm: this.createTechRecordForm.get('vrm').value,
+      vType: this.createTechRecordForm.get('vehicleType').value
```
rspasici: it will look cleaner if you move those value getters into a separate object and dispatch that object as payload

src/app/technical-record-create/technical-record-create.component.ts
```diff
@@ -7,10 +14,37 @@ import { Component, OnInit, ChangeDetectionStrategy } from '@angular/core';
...
+  vehicleTypes = ['PSV', 'HGV', 'Trailer'];
+  createTechRecordForm: FormGroup;
 
-  constructor() { }
+  constructor(private _store: Store<IAppState>) {
+  }
 
   ngOnInit() {
+    this.createTechRecordForm = new FormGroup({
+      'vehicleType': new FormControl('PSV', Validators.required),
```
m-ujaffer: we can make a interface called `CreateTechRecordVM` which has the following attribute:
```
interface CreateTechRecordVM{
vin: string,
vrm: string,
vType:string
}
```
This can then be used in the store dispatch and in the action and reducer

PR #91, opened 2020-02-25
feature/CVSB-11329 - [FE] As the DVSA, we want to be able to search for vehicles when the full VIN is less than 9 characters in VTM

irina005:
https://jira.dvsacloud.uk/browse/CVSB-11329

### DIFF
--- a/src/app/pipes/DefaultNullOrEmptyPipe.ts
+++ b/src/app/pipes/DefaultNullOrEmptyPipe.ts
@@ -8,22 +8,23 @@ export class DefaultNullOrEmpty implements PipeTransform {
     switch (typeof value) {
       case 'string':
         if (value.trim().length > 0) {
-
-          if (!/\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z/.test(value)) {
+          if (/\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z/.test(value) || /(\d{4})-(\d{2})-(\d{2})/.test(value)) {
+            const checkDate = new Date(value);
+            if (checkDate.toISOString() === value) {
+              return checkDate.getDate() + '/' +  ('0' + (checkDate.getMonth() + 1)).slice(-2) + '/' + checkDate.getFullYear();
+            } else {
+              const dateArray = value.split('-');
+              return dateArray[2] + '/' + dateArray[1] + '/' + dateArray[0];
+            }
+          } else {
             return value;
           }
-          const checkDate = new Date(value);
-
-          if (checkDate.toISOString() === value) {
-            return checkDate.getDate() + '/' + checkDate.getMonth() + '/' + checkDate.getFullYear();
-          }
-
         } else {
           return '-';
         }
         break;
       case 'boolean':
-        return (value == null) ? '-' : (value ? 'Yes' : 'No');
+        return value ? 'YES' : 'NO';
         break;
       default:
         return (value == null) ? '-' : value;
--- a/src/app/store/effects/VehicleTestResultModel.effects.spec.ts
+++ b/src/app/store/effects/VehicleTestResultModel.effects.spec.ts
@@ -1,0 +1,91 @@
+import {Observable, of, ReplaySubject} from 'rxjs';
+import {TechnicalRecordServiceMock} from '../../../../testconfig/services-mocks/technical-record-service.mock';
+import {TestBed} from '@angular/core/testing';
+import {TechnicalRecordService} from '@app/technical-record-search/technical-record.service';
+import {RouterTestingModule} from '@angular/router/testing';
+import {INITIAL_STATE, Store} from '@ngrx/store';
+import {hot} from 'jasmine-marbles';
+import {provideMockActions} from '@ngrx/effects/testing';
+import {EVehicleTestResultModelActions,
+        GetVehicleTestResultModel,
+        GetVehicleTestResultModelFailure,
+        GetVehicleTestResultModelSuccess} from '@app/store/actions/VehicleTestResultModel.actions';
+import {IAppState} from '@app/store/state/app.state';
+import {VehicleTestResultModelEffects} from '@app/store/effects/VehicleTestResultModel.effects';
+import {TestResultService} from '@app/technical-record-search/test-result.service';
+import {TestResultModel} from '@app/models/test-result.model';
+
+const testResult: TestResultModel = {
+  testerStaffId: '1234',
+  testStartTimestamp: new Date('2019-01-14T10:36:33.987Z'),
+  odometerReadingUnits: 'string',
+  testEndTimestamp: new Date('2019-01-14T10:36:33.987Z'),
+  testStatus: 'string',
+  testTypes: [],
+  vehicleClass: null,
+  vin: '1234',
+  vehicleSize: 'string',
+  testStationName: 'string',
+  vehicleId: '3121',
+  noOfAxles: 2,
+  vehicleType: 'string',
+  countryOfRegistration: 'string',
+  preparerId: '2332',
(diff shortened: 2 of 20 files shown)


src/app/technical-record-search/technical-record.service.ts
```diff
@@ -14,17 +14,16 @@ export class TechnicalRecordService {
   constructor(private httpClient: HttpClient,
     private _appConfig: AppConfig) {
     this.routes = {
-      // techRecords: (searchIdentifier: string) => `${this._apiServer.APITechnicalRecordServerUri}/vehicles/${searchIdentifier}/tech-records`,
-      techRecordsAllStatuses: (searchIdentifier: string) => `${this._apiServer.APITechnicalRecordServerUri}/vehicles/${searchIdentifier}/tech-records?status=all&metadata=true`,
+      techRecordsAllStatuses: (searchIdentifier: string, searchCriteria: string) => `${this._apiServer.APITechnicalRecordServerUri}/vehicles/${searchIdentifier}/tech-records?status=all&metadata=true&searchCriteria=${searchCriteria}`,
```
m-ujaffer: break into to two lines.

src/app/technical-record-search/technical-record-search.component.ts
```diff
@@ -15,15 +17,39 @@ export class TechnicalRecordSearchComponent {
...
-  public searchTechRecords(q: string) {
+  public searchTechRecords(searchIdentifier: string, searchCriteria: string) {
+
     this.isLoading = true;
-    this.searchIdentifier = q;
-    this._store.dispatch(new GetVehicleTechRecordModelHavingStatusAll(q));
+    this.searchIdentifier = searchIdentifier;
+
+    switch (searchCriteria) {
+      case SEARCH_CRITERIA.ALL_CRITERIA:
```
m-ujaffer: remove this and move to default.

src/app/technical-record-search/technical-record-search.component.ts
```diff
@@ -15,15 +17,39 @@ export class TechnicalRecordSearchComponent {
...
+      case SEARCH_CRITERIA.FULL_VIN_CRITERIA:
+        this.searchCriteria = 'vin';
+        break;
+      case SEARCH_CRITERIA.PARTIAL_VIN_CRITERIA:
+        this.searchCriteria = 'partialVin';
+        break;
+      case SEARCH_CRITERIA.TRL_CRITERIA:
+        this.searchCriteria = 'trailerId';
+        break;
+      default:
```
m-ujaffer: change to following
```
      case SEARCH_CRITERIA.ALL_CRITERIA:
      default:
        this.searchCriteria = 'all';
```

src/app/technical-record-search/technical-record.service.ts
m-ujaffer: The default value for searchCriteria is 'all'. You don't need to pass this parameter if you are always searching for all.

src/app/technical-record-search/technical-record-search.component.html
```diff
@@ -25,14 +27,22 @@ <h1 class="govuk-heading-xl">Search for a technical record</h1>
...
+            Search criteria
+          </label>
+          <select class="govuk-select" id="test-search-criteria" #searchCriteria>
+            <option *ngFor="let searchCriteria of searchCriteriaOptions" [value]="searchCriteria">{{ searchCriteria }}</option>
+          </select>
+        </div>
 
-        <a role="button" (click)="searchTechRecords(q.value)" draggable="false"
-          class="govuk-button govuk-button--start mt-1" data-module="govuk-button">
+        <a role="button" (click)="searchTechRecords(searchIdentifier.value, searchCriteria.value)" draggable="false" class="govuk-button govuk-button--start mt-1"
```
alleypa: no custom CSS please follow this link https://design-system.service.gov.uk/styles/spacing/

src/app/pipes/DefaultNullOrEmptyPipe.ts
```diff
@@ -8,22 +8,23 @@ export class DefaultNullOrEmpty implements PipeTransform {
     switch (typeof value) {
       case 'string':
         if (value.trim().length > 0) {
-
-          if (!/\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z/.test(value)) {
+          if (/\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z/.test(value) || /(\d{4})-(\d{2})-(\d{2})/.test(value)) {
```
alleypa: Can we not use angular date pipe  for this implementation. Hence no need for this code and associated test

src/app/store/effects/VehicleTestResultModel.effects.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+import {EVehicleTestResultModelActions,
+        GetVehicleTestResultModel,
+        GetVehicleTestResultModelFailure,
+        GetVehicleTestResultModelSuccess} from '@app/store/actions/VehicleTestResultModel.actions';
+import {IAppState} from '@app/store/state/app.state';
+import {VehicleTestResultModelEffects} from '@app/store/effects/VehicleTestResultModel.effects';
+import {TestResultService} from '@app/technical-record-search/test-result.service';
+import {TestResultModel} from '@app/models/test-result.model';
+
+const testResult: TestResultModel = {
```
alleypa: For testing we need this much data. Please lets discuss the approach

src/app/technical-record/adr-details/adr-details-form/store/adrDetails.reducer.spec.ts
```diff
@@ -46,6 +46,15 @@ describe('adrDetailsReducer', () => {
       expect(result.additionalNotesOptions.length).toBe(1);
       expect(result.additionalNotesOptions[0]).toBe('12');
     });
+
```
alleypa: this old adr-details files will be deleted

src/app/pipes/DefaultNullOrEmptyPipe.ts
irina005: I did this pipe because for each empty or null field (being it string/date string/boolean, etc) we have to display hyphen. If we use date pipe we have to do the empty/null check for each field in the html template. This was an idea of optimisation I discussed with @m-ujaffer

src/app/store/effects/VehicleTestResultModel.effects.spec.ts
irina005: I added more data because the error on the tests failing specifically said that data for field x from TestResultModel was missing. Sure, we can discuss any approach.

src/app/technical-record/adr-details/adr-details-form/store/adrDetails.reducer.spec.ts
irina005: I know what you are saying. We had a chat regarding this some time ago. I added this to improve coverage on the app. This can be removed when you are going to merge the ADR work into develop.

src/app/pipes/DefaultNullOrEmptyPipe.ts
alleypa: Irina, checking for an empty/null DATE filed and using a date pipe is still a much simpler code than producing lots of re-engineered code. I see no benefit. I agree with other string/null fields concept

src/app/store/effects/VehicleTestResultModel.effects.spec.ts
alleypa: yes there is a simpler approach. for example `const testResult = {} as TestReult`. will work just fine. You only need to show the intentions here and not the either payload

alleypa: show few keys with that approach.

PR #92, opened 2020-03-02
Fxied test result showing error when not found

rspasici:
https://jira.dvsacloud.uk/browse/CVSB-12578

### DIFF
--- a/src/app/shared/delayed-retry/delayed-retry.ts
+++ b/src/app/shared/delayed-retry/delayed-retry.ts
@@ -1,12 +1,12 @@
 import { Observable, of, throwError } from 'rxjs';
 import { delay, mergeMap, retryWhen, switchMap } from 'rxjs/operators';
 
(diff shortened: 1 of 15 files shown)


src/app/shared/delayed-retry/delayed-retry.ts
```diff
@@ -1,12 +1,12 @@
 import { Observable, of, throwError } from 'rxjs';
 import { delay, mergeMap, retryWhen, switchMap } from 'rxjs/operators';
 
-const getErrorMessage = { error: `Unable to load data. Please contact the help desk or refresh the page to try again.`};
+export const getErrorMessage = { error: `Unable to load data. Please contact the help desk or refresh the page to try again.` };
```
m-ujaffer: The export is never used

m-ujaffer: ideally should be in enum if needs to be used in more than one place.

rspasici: this is a specific error only thrown by this method. removing the export

PR #107, opened 2020-03-30
feature/CVSB-10283 - CVSB-10283 work + last changes requested:

irina005:
- created test-record container
- created selector to grab the test result and test type by testTypeNumber
- changed testTypes applicable arrays to be objects containing test types names
- did changes accordingly into test specs and the other test section components

Update:
- rebased with common
- removed styles from components & used the global.css [changed the order of loading from angular.json]
- added ids for testers 
- updated the models
- improved the logic in some templates (i.e. test-record, defects) 
- fixed tests failing due to the changes


src/app/models/test.type.ts
```diff
@@ -21,4 +26,13 @@ export interface TestType {
   name: string;
   certificateLink: string;
   testResult: string;
+  testExpiryDate: Date;
```
alleypa: Type should be string. this is what Backend sends 'YYYY-MM-DD'

irina005: image

irina005: @alleypa not always

alleypa: that fine Irina, I think better to still treat it as string. The conversion will just fine

src/app/shared/libraries/accordion-item/accordion-item.component.ts
```diff
@@ -0,0 +1,22 @@
...
+@Component({
+  selector: 'vtm-accordion-item',
+  templateUrl: './accordion-item.component.html',
+  styleUrls: ['./accordion-item.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class AccordionItemComponent implements OnInit {
+
+  @Input() title: string;
+  // @Input() hide: boolean;
```
alleypa: you can remove the commented codes and the ngOnint

src/app/shared/shared.module.ts
```diff
@@ -1,17 +1,17 @@
...
-import { FilterRecordPipe } from '@app/pipes/FilterRecordPipe';
-import { KeysPipe } from '@app/pipes/keysPipe';
-import { OrderByStatusPipe } from '@app/pipes/OrderByStatusPipe';
-import { IsPrimaryVrmPipe } from '../pipes/IsPrimaryVrmPipe';
-import { AdrReasonModalComponent } from './adr-reason-modal/adr-reason-modal.component';
-import { ErrorSummaryComponent } from './error-summary/error-summary.component';
-import { PageNotFoundComponentComponent } from './page-not-found-component/page-not-found-component.component';
-import { PendingChangesService } from './pending-changes-service/pending-changes.service';
-import { PreventLeavePageModalComponent } from './prevent-page-leave-modal/prevent-leave-page-modal.component';
+import {CommonModule} from '@angular/common';
```
alleypa: introduce prettier of well structure

src/app/test-record/defects/defects.component.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+  let fixture: ComponentFixture<DefectsComponent>;
+  const testType = {} as TestType;
+  const testRecord = {} as TestResultModel;
+
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      imports: [ SharedModule ],
+      declarations: [ DefectsComponent ],
+      providers: [
+        TechRecordHelpersService
```
alleypa: please clean up. Only use all that is necessary

src/app/test-record/defects/defects.component.ts
```diff
@@ -0,0 +1,22 @@
...
+import {TestType} from '@app/models/test.type';
+import {TestResultModel} from '@app/models/test-result.model';
+
+@Component({
+  selector: 'vtm-defects',
+  templateUrl: './defects.component.html',
+  styleUrls: ['./defects.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class DefectsComponent implements OnInit {
```
alleypa: you should be able to write test to mock these @Input(). Please check the ADR branch or lets talk

src/app/test-record/defects/defects.component.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  @Input() testType: TestType;
```
alleypa: you should be able to write test to mock these @Input(). Please check the ADR branch or lets talk

src/app/test-record/notes/notes.component.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+
+describe('NotesComponent', () => {
+  let component: NotesComponent;
+  let fixture: ComponentFixture<NotesComponent>;
+  const testType = {} as TestType;
+
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      imports: [ SharedModule ],
+      providers: [
```
alleypa: what are achieving with TechRecordHelpersService. Less code. Only necessary code

src/app/test-record/seatbelt-installation-check/seatbelt-installation-check.component.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+describe('SeatbeltInstallationCheckComponent', () => {
+  let component: SeatbeltInstallationCheckComponent;
+  let fixture: ComponentFixture<SeatbeltInstallationCheckComponent>;
+  const testType = {} as TestType;
+
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      imports: [ SharedModule ],
+      providers: [
+        TechRecordHelpersService
```
alleypa: confusion with the use of TechRecordHelpersService??

src/app/test-record/defects/defects.component.ts
irina005: image

irina005: image

src/app/test-record/test-record.mapper.ts
```diff
@@ -0,0 +1,56 @@
+import { Injectable } from '@angular/core';
```
alleypa: well done

package.json
```diff
@@ -81,6 +81,7 @@
     "jest-plugin-context": "^2.9.0",
     "jest-runner-tsc": "^1.6.0",
     "jest-sonar-reporter": "^2.0.0",
+    "prettier": "2.0.2",
```
FiF0o, line 84: I am confused why we have prettier changes in this PR. I don't think the ticket is mentioning it. Also, will these formatting changes impact the existing linting (and CI)?. It might be worth adding a command or a to a hook for it. Why still using the .editorconfig ? Prettier should format the code for us before committing. I can see a max line of length of 180 whereas prettier has 98 for example.

m-ujaffer: Prettier was added recently primarily for code formatting as it is better at it than ESLint in regards to code formatting. . It was actually planned to go with the ADR changes but it is added over here as well. ADR would go into Develop first so this would end up conflicting. We can remove it from this branch and keep it with ADR.

alleypa: @JonLazariniDD having prettier in this ticket does not cause much of an issue at this stage. Yes it was added and needed as part of the huge ADR refactor ticket. The VTM app currently lacks formatting and proper code structuring. The developer added it here to start conforming to the same code layout and structure before we eventually merge ADR to develop. Hence once ADR goes to develop this wont be a problem. So I suggest we leave it for now now. With respect to .editorConfig this is a default with most Angular setup project. We are using Prettier going forward and it supersedes .editorConfig

JonLazariniDD: ok, are we sure it is not going to cause issues with SonarQube? I tried to check the log and seen that the CI env does not have the correct version of Ts installed 2.9.2 whereas the CI is looking for 3.x. which is a different issue.
```
ERROR: You are using version of TypeScript 2.9.2 which is not supported; supported versions >=3.2.1
```
I can see that sonarQube is expecting a linting report to be analysed
https://github.com/dvsa/cvs-app-vtm/blob/develop/sonar-project.properties#L18
but it seems we are never generating that so I assume the CI won't pick this up?
I don't have access to SonarQube yet to verify the generated report unfortunately.

alleypa: @JonLazariniDD you hit the nail right on the head. From my point of view and the analysis so far, we are in somewhat of a virgin island. So as much of the all the necessary good stuff and best practices can be pushed through for now. The VTM app needs a lot of TLC. With respect to typescript version - this is another deficiency with the VTM app. Ideally we should be advising for an upgrade to ng9

src/app/test-record/defects/defects.component.scss
```diff
@@ -0,0 +1,25 @@
+
+.govuk-summary-list {
+
+  .tag-custom {
+    padding: 1px 10px 1px 10px;
+    vertical-align: middle;
+  }
+
+  & .tag--black {
+    background-color: #0b0c0c !important;
```
JonLazariniDD, line 10: Is there a way we avoid this? this will make overrides impossible if we have to extend it and also it would probably nice to use a theme variable for the scss background-color, it will make it easier to maintain later.
It is probably worth thinking about a way to scale the scss and architect it.
I think atomic design pattern could be an approach maybe? Thoughts

src/app/test-record/defects/defects.component.scss
```diff
@@ -0,0 +1,25 @@
...
+    @extend .tag-custom;
+  }
+
+  & .tag--red {
+    background-color: #d4351c !important;
```
JonLazariniDD, line 15: same here for the background-color variable.

src/.prettierrc
```diff
@@ -0,0 +1,11 @@
+{
```
JonLazariniDD, line 1: Why have we got this file in the PR? It does not seem part of the ticket.

src/app/models/test.type.ts
```diff
@@ -21,4 +26,13 @@ export interface TestType {
   name: string;
   certificateLink: string;
   testResult: string;
+  testExpiryDate: string;
+  testAnniversaryDate: Date;
```
JonLazariniDD, line 30: I guess this is addressed as part of 10282 and will be a string

src/app/shared/libraries/accordion-item/accordion-item.component.html
```diff
@@ -0,0 +1,15 @@
+
+<div class="govuk-accordion__section">
+  <div class="govuk-accordion__section-header">
+    <h2 class="govuk-accordion__section-heading">
+      <span class="govuk-link govuk-!-font-size-36">{{title}}</span>
+      <span class="govuk-accordion__section-button" id="accordion-default-heading"></span>
+    </h2>
+  </div>
+  <div id="accordion-default-content" class="govuk-accordion__section-content">
```
JonLazariniDD, line 9: i assume we need an id for enabling the expand/collapse logic. This id seems quite generic, can we have a naming more unique assuming we can have multiple please? We might need to dynamically create the id if it is the case, it could be the section name for example `accordion-default-content${titleSection}` (titleSection will be the unique name/section of the form)

package.json
irina005: Closing it because is outdated

src/.prettierrc
irina005: It is in due to develop rebasing after ADR merge.

irina005:
Closing this PR in order to point it to common branch.

PR #109, opened 2020-03-30
ADR CVSB-10083 & 10081:

alleypa:
https://jira.dvsacloud.uk/browse/CVSB-10083
https://jira.dvsacloud.uk/browse/CVSB-10081

  - Add main/sub component implemenation to display ADR details in readonly mode
  - Add features to enable ADR editing
  - Add implementation to guard ADR removal
  - Add implementation to cancel ADR editing
  - Add effect/action to save edited ADR - TechRecord Details
  - Add implementation to display multiple records select screen
  - Remove <any> and added correct types everywhere in the app
  - Add upload/download functionality
  - Add unit testing to functionalities


alleypa:
> I started to have a look at the code, the squash commit history is nice as it is grouped into logical commit and will give a clean history once merged.
> It seems that we are mixing a feature and the introduction of new tooling, config with prettier as well as some linting fixes.
> I think we need to separate them into 2 'things', one having the desired features and another one for the config as this will potentially affect the configuration of SonarQube (and existing rules in place?) or at least that is something we will have to consider and we might even have to run some regression depending on what is picked in the CI.
> Also if for some reasons we had to revert back, it would be nice to keep the prettier config for example.

alleypa:
Dropping it for testers to get updates

PR #124, opened 2020-04-10
Feature/CVSB-14184 - Issue when changing adr details from yes to no without actually saving the change

irina005:
https://jira.dvsacloud.uk/browse/CVSB-14184

### DIFF
--- a/src/app/shared/adr-reason-modal/adr-reason-modal.component.ts
+++ b/src/app/shared/adr-reason-modal/adr-reason-modal.component.ts
@@ -14,9 +14,14 @@ export interface DialogData {
 export class AdrReasonModalComponent {
   constructor(
     public dialogRef: MatDialogRef<AdrReasonModalComponent>,
-    @Inject(MAT_DIALOG_DATA) public data: DialogData) { }
+    @Inject(MAT_DIALOG_DATA) public data: DialogData
+  ) {}
 
   close(): void {
-    this.dialogRef.close();
+    this.dialogRef.close(false);
+  }
+
+  save(modalData:  string): void {
+    this.dialogRef.close({isSave: true, data: modalData});
   }
 }
--- a/src/app/store/effects/VehicleTechRecordModel.effects.ts
+++ b/src/app/store/effects/VehicleTechRecordModel.effects.ts
@@ -180,8 +180,7 @@ export class VehicleTechRecordModelEffects {
                * error not working properly yet
                */
               const errorMessage = error[0].message;
-              this._store.dispatch(new SetErrorMessage([errorMessage]));
-              return [];
+              return [new SetErrorMessage([errorMessage]), new SetViewState(VIEW_STATE.VIEW_ONLY)];
             })
           );
       }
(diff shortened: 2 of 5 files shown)


src/app/shared/adr-reason-modal/adr-reason-modal.component.ts
```diff
@@ -14,9 +14,14 @@ export interface DialogData {
...
+    @Inject(MAT_DIALOG_DATA) public data: DialogData
+  ) {}
 
   close(): void {
-    this.dialogRef.close();
+    this.dialogRef.close(false);
+  }
+
+  save(modalData:  string): void {
+    this.dialogRef.close({save: true, data: modalData});
```
alleypa, line 25: use isSave

src/app/store/effects/VehicleTechRecordModel.effects.ts
```diff
@@ -180,8 +180,7 @@ export class VehicleTechRecordModelEffects {
                * error not working properly yet
                */
               const errorMessage = error[0].message;
-              this._store.dispatch(new SetErrorMessage([errorMessage]));
-              return [];
+              return [new SetErrorMessage([errorMessage]), new SetViewState(VIEW_STATE.VIEW_ONLY)];
```
alleypa, line 183: why do we need to dispatch SetViewState(VIEW_STATE.VIEW_ONLY)?

src/app/technical-record/technical-record.component.ts
```diff
@@ -109,9 +109,9 @@ export class TechnicalRecordComponent implements OnInit {
         }
       });
 
-      dialogRef.afterClosed().subscribe((reasonForChanges) => {
-        if (reasonForChanges) {
-          mergedRecord.reasonForCreation = reasonForChanges;
+      dialogRef.afterClosed().subscribe((result) => {
+        if (!!result.save) {
```
alleypa, line 113: use - if (result && result.isSave)

src/app/shared/adr-reason-modal/adr-reason-modal.component.ts
irina005: done

src/app/store/effects/VehicleTechRecordModel.effects.ts
irina005: The bug is detailed in the ticket. If the backend PUT returned an error and the user then clicked CANCEL the state was not updated to view.

alleypa: All you need to do here is `return [new SetErrorMessage([errorMessage]);`. The CVSB-14184 bug refers to the `Yes` or `No` of the ADR details not working. That bug in itself is not reproducible because the implementation is already in the VTM-App. Except I am missing something else.  

To clear summary errors when cancel link is clicked please create an effect `clearErrorMessage$` in the error.effects.ts. The effect should have `ofType<SetViewState>` then filter where `VIEW_STATE===VIEW_ONLY`. The final step should be `map(() => ClearErrorMessage()`. This will layout the ground work for error-handling in the VTM-APP at large. More to come when validation ticket is played

irina005: Let's have a call tomorrow with Dragos and discuss why you don't consider it as bug and why he does. Once clarified, I can make whichever amend is necessary.

alleypa:
solution to be as an update in 10597 branch

PR #126, opened 2020-04-15
feature/CVSB-10282 - Test Record Edit Validation

irina005:
Ticket:
https://jira.dvsacloud.uk/browse/CVSB-10282


sonar-project.properties
```diff
@@ -13,7 +13,7 @@ sonar.sourceEncoding=UTF-8
 # Path is relative to the sonar-project.properties file. Replace “\” by “/” on Windows.
 # This property is optional if sonar.modules is set.
 sonar.sources=src/app
-sonar.exclusions=test-config/**, cypress/**, dist/**, .nyc_output/**, .scannerwork/**, coverage/**, **/*.module.ts, **/*.routes.ts, **/*.state.ts, **/utils.ts
+sonar.exclusions=test-config/**, cypress/**, dist/**, .nyc_output/**, .scannerwork/**, coverage/**, **/*.module.ts, **/*.routes.ts, **/*.state.ts, **/utils.ts, **/*.utils.ts
```
alleypa, line 16: why this change?

src/app/models/test.type.ts
```diff
@@ -9,7 +14,7 @@ export interface TestType {
   numberOfSeatbeltsFitted: number;
   testTypeEndTimestamp: Date;
   reasonForAbandoning: string;
-  lastSeatbeltInstallationCheckDate: string;
+  lastSeatbeltInstallationCheckDate: Date;
```
alleypa, line 17: the API spec specifies this Date fields as string. Please lets obey that. Change where necessary to use string instead of the date type

src/app/shared/components/autocomplete/autocomplete.component.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    fixture.componentInstance.testAutocompleteControl = new FormControl('testing');
+    fixture.detectChanges();
+    tick();
+
+    expect(inputElement.nativeElement.value).toEqual('testing');
+  }));
+
+});
+
+@Component({
```
alleypa, line 50: You test wrapper should just be a simple like this `<div [formControl]="testAutocompleteControl"></div>` . The test is to show how this control will work and not from the <vtm-date-input> perpective

src/app/shared/dialog-box/dialog-box.component.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    });
+  });
+
+  describe('save', () => {
+    it('should close the modal & send modal data', () => {
+      const modalData = { isSave: true, data: 'test data' };
+      spyOn(component.dialogRef, 'close');
+      component.save('test data');
+      expect(component.dialogRef.close).toHaveBeenCalled();
```
alleypa, line 50: you dont need line 50

src/app/shared/dialog-box/dialog-box.component.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+  it('should create', () => {
+    expect(component).toBeTruthy();
+    expect(fixture).toMatchSnapshot();
+  });
+
+  describe('close', () => {
+    it('should close the modal when called', () => {
+      spyOn(component.dialogRef, 'close');
+      component.close();
+      expect(component.dialogRef.close).toHaveBeenCalled();
```
alleypa, line 39: you need line 39

src/app/shared/libraries/accordion-item/accordion-item.component.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+    .compileComponents();
+  }));
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(AccordionItemComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
```
alleypa, line 22: we also need a test to show content passed to the accordionItem are also rendered

src/app/shared/libraries/accordion/accordion.component.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+    .compileComponents();
+  }));
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(AccordionComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
```
alleypa, line 22: likewise one more test here

src/app/test-record/vehicle/vehicle-edit/vehicle-edit.component.html
```diff
@@ -0,0 +1,184 @@
+<div class="govuk-grid-row">
```
alleypa, line 1: To check whether to display errors you should be checking for the valid or invalid status of the particular field NOT like this `[class.govuk-form-group--error]="formErrors?.includes('Enter the vin')`

sonar-project.properties
irina005: As per one of our discussions, I added the utils/test.utils to be ignored by sonar. Please let me know if you changed your mind.

src/app/test-record/vehicle/vehicle-edit/vehicle-edit.component.html
irina005: I have a function in test-record component that checks the **validity** of **each** field in the form and adds the error accordingly. Would you please check that? Thanks.

src/app/models/test.type.ts
irina005: It is confusing because some of them have already **Date** as type. Look at **testTypeEndTimestamp** above. I agree that we should be consistent and I will change. Thanks for noticing

src/app/shared/dialog-box/dialog-box.component.spec.ts
irina005: Please give me some clarification. The **save** function triggers the closing of the modal. (you can check the ts file) Why not testing if closed was called?

src/app/test-record/vehicle/vehicle-edit/vehicle-edit.component.html
```diff
@@ -0,0 +1,184 @@
...
+  <div class="govuk-grid-column-three-quarters">
+    <div class="govuk-form-group">
+      <fieldset class="govuk-fieldset">
+        <div class="govuk-form-group">
+          <div class="govuk-grid-row govuk-!-margin-bottom-3" [class.govuk-form-group--error]="formErrors?.includes('Enter the vin')">
+            <div class="govuk-grid-column-one-half">
+              <label class="govuk-label govuk-!-font-weight-bold">
+                Vehicle identification number (VIN)
+              </label>
+              <span *ngIf="formErrors?.includes('Enter the vin')" class="govuk-error-message">
```
alleypa, line 11: <span id="your-error-id"
                          class="govuk-error-message" *ngIf="testResultChildForm.get('vin').invalid">
                <span class="govuk-visually-hidden">Error:</span>
                <span>Enter the VIN</span>
              </span>

sonar-project.properties
alleypa: it doesn't need to be sonar ignored. the file is used sufficiently in unit tests hence should not affect coverage

src/app/shared/dialog-box/dialog-box.component.spec.ts
alleypa: line 51 does the same thing

src/app/test-record/test-record.component.spec.ts
```diff
@@ -1,55 +1,115 @@
...
+      schemas: [CUSTOM_ELEMENTS_SCHEMA],
       providers: [
+        TestRecordMapper,
+        FormGroupDirective,
         {
           provide: Store,
           useValue: {
             dispatch: jest.fn(),
-            pipe: jest.fn(() => hot('-a', {a: INITIAL_STATE})),
+            pipe: jest.fn(() => hot('-a', { a: INITIAL_STATE })),
```
alleypa, line 46: we should remove store references in the component

src/app/test-record/test-record.component.spec.ts
```diff
@@ -1,55 +1,115 @@
-import {async, ComponentFixture, getTestBed, TestBed} from '@angular/core/testing';
-
-import {TestRecordComponent} from './test-record.component';
-import {INITIAL_STATE, Store} from '@ngrx/store';
-import {IAppState} from '@app/store/state/app.state';
-import {hot} from 'jasmine-marbles';
-import {ActivatedRoute, convertToParamMap} from '@angular/router';
-import {Observable, of} from 'rxjs';
+import { async, ComponentFixture, getTestBed, TestBed } from '@angular/core/testing';
```
alleypa, line 1: some imports are not needed in this file

src/app/test-record/defects/defects.component.html
```diff
@@ -0,0 +1,75 @@
...
+      </dt>
+      <dd class="govuk-summary-list__value">
+        {{ ((defect.deficiencyId ? defect.deficiencyId : '') + ' ' + (defect.deficiencySubId ? defect.deficiencySubId : '') + ' ' + (defect.deficiencyText ? defect.deficiencyText : '')) | DefaultNullOrEmpty}}
+      </dd>
+    </div>
+    <div class="govuk-summary-list__row">
+      <dt class="govuk-summary-list__key">
+        Location
+      </dt>
+      <dd class="govuk-summary-list__value">
```
alleypa, line 39: is it possible to move this logic to the component side?

src/app/technical-record/tyres/tyres.component.spec.ts
```diff
@@ -1,64 +1,67 @@
...
-      'vin': 'XMGDE02FS0H012345',
-      'vehicleSize': 'small',
-      'testStationName': 'Rowe, Wunsch and Wisoky',
-      'vehicleId': 'JY58FPP',
-      'vehicleType': 'psv',
-      'axles': [
-        { 'parkingBrakeMrk': false, 'axleNumber': 1 },
-        { 'parkingBrakeMrk': true, 'axleNumber': 2 },
-        { 'parkingBrakeMrk': false, 'axleNumber': 3 }
+      vin: 'XMGDE02FS0H012345',
```
alleypa, line 25: we should have a smaller mock object, Within each test partially mutate the key you want to update. I can explain better

src/app/store/state/ReferenceDataState.state.ts
```diff
@@ -0,0 +1,12 @@
+import { Preparer } from '@app/models/preparer';
+import { TestStation } from '@app/models/test-station';
+
+export interface IReferenceDataState {
```
alleypa, line 4: ReferenceDataState instead of IReferenceDataState

sonar-project.properties
irina005: Sonar included the file. I can show you. We can ask Umer for his oppinion as well.

src/app/test-record/test-record.component.spec.ts
irina005: I tried removing as we already talked about this and the test was failing

src/app/shared/directives/digit-only.directive.ts
```diff
@@ -0,0 +1,17 @@
+import {Directive, HostListener} from '@angular/core';
+
+@Directive({
+  selector: '[appDigitOnly]'
```
alleypa, line 4: rename to vtmDigitOnly

src/app/shared/directives/digit-only.directive.spec.ts
```diff
@@ -0,0 +1,16 @@
+import { DigitOnlyDirective } from './digit-only.directive';
+
+describe('DigitOnlyDirective', () => {
+  it('should create an instance', () => {
+    const directive = new DigitOnlyDirective();
+    expect(directive).toBeTruthy();
+  });
+  it('should listen to KeyboardEvent', () => {
```
alleypa, line 8: Rename test to `should call prevent default on non numeric key press` and you can elaborate on your event object like this
const event: KeyboardEvent = {
 key: a,
 preventDefault: () => {}
}

So you can check for both numeric and and non-numeric parsing in the test. So likely one more test for numeric

src/app/shared/components/autocomplete/autocomplete.component.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+  it('should populate the internal input when an external value is provided', fakeAsync(() => {
+    testFixture.componentInstance.testAutocompleteControl = new FormControl('testing');
+    testFixture.detectChanges();
+    tick();
+
+    expect(inputElement.nativeElement.value).toEqual('testing');
+    expect(testFixture).toMatchSnapshot();
+  }));
+
+  it('should correctly render the passed @Input value', () => {
```
alleypa, line 57: I think you dont need this test

src/app/shared/components/autocomplete/autocomplete.component.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+    fixture.detectChanges();
+    expect(component._value).toBe('test input');
+  });
+
+  it('should correctly write value into input field', () => {
```
alleypa, line 62: also you dont need this test. You doing it already above

src/app/store/actions/ReferenceData.actions.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+import {
+  EReferenceDataActions,
+  LoadPreparers,
+  LoadPreparersSuccess,
+  LoadTestStations,
+  LoadTestStationsSuccess
+} from '@app/store/actions/ReferenceData.actions';
+
+describe('LoadPreparers', () => {
+  test('action should have correct type', () => {
```
alleypa, line 12: we should be consistent with the use of `it` instead of `test`

src/app/store/reducers/ReferenceData.reducers.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+
+      expect(result).toMatchObject({
+        ...initialReferenceDataState,
+        testStations: {} as TestStation[]
+      });
+    });
+  });
+
+  describe('should return default if action is different', () => {
+    test('return current state value', () => {
```
alleypa, line 36: `should return current state`

src/app/store/reducers/ReferenceData.reducers.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+      const result = ReferenceDataReducers(initialReferenceDataState, action);
+
+      expect(result).toMatchObject({
+        ...initialReferenceDataState,
+        testStations: {} as TestStation[]
+      });
+    });
+  });
+
+  describe('should return default if action is different', () => {
```
alleypa, line 35: we can shorten the describe title `LoadTestStations`

src/assets/config/config.dev.json
```diff
@@ -18,7 +17,7 @@
     "clientId": "",
     "tenant": "",
     "redirectUri": "http://localhost:4200",
-    "cacheLocation": "localStorage",
+    "cacheLocation": "",
```
alleypa, line 20: undo this

src/assets/config/config.dev.json
```diff
@@ -1,9 +1,8 @@
-{
-  "env": {
-    "name": "dev"
-  },
+{  "env": {
+  "name": "dev"
+},
   "appInsights": {
-    "instrumentationKey": "#{appInsightsKey}"
+    "instrumentationKey": ""
```
alleypa, line 5: undo this

src/app/test-record/visit/visit.component.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+import { SharedModule } from '@app/shared/shared.module';
+import { TestResultModel } from '@app/models/test-result.model';
+import { Component, Input } from '@angular/core';
+import { VIEW_STATE } from '@app/app.enums';
+import { TestStation } from '@app/models/test-station';
+
+describe('VisitComponent', () => {
+  let component: VisitComponent;
+  let fixture: ComponentFixture<VisitComponent>;
+  const testRecord = {} as TestResultModel;
```
alleypa, line 13: try to elaborate on few members of this type

src/app/test-record/test-record.component.ts
```diff
@@ -14,32 +24,103 @@ import {TestType} from '@app/models/test.type';
...
+
+  onFormChanges(): void {
+    this.testResultParentForm.valueChanges.subscribe((val) => {
+      this.isFormDirty.emit(true);
     });
   }
 
+  getInvalidControls(form: FormGroup): string[] {
+    const invalidControls: string[] = [];
+    const checkInvalid = (formtoCheck: FormGroup) => {
```
alleypa, line 110: you separate this to a standalone function

src/app/test-record/test-record.container.spec.ts
```diff
@@ -0,0 +1,103 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
```
alleypa, line 1: check ADR branch on container testing. Let me know I'll explain more

src/app/test-record/test-record.component.spec.ts
```diff
@@ -1,55 +1,128 @@
...
+    spyOn(component.switchState, 'emit');
+    component.switchCurrentState('view');
+    expect(component.switchState.emit).toHaveBeenCalled();
+    expect(component.switchState.emit).toHaveBeenCalledWith(VIEW_STATE.VIEW_ONLY);
   });
 
+  it('should switch current state to edit', () => {
+    spyOn(component.switchState, 'emit');
+    component.switchCurrentState('edit');
+    expect(component.switchState.emit).toHaveBeenCalled();
```
alleypa, line 98: you only need toHaveBeenCalledWith...

src/app/test-record/test-record.component.spec.ts
```diff
@@ -1,55 +1,128 @@
...
+    expect(component.onFormChanges).toHaveBeenCalled();
+  });
+
+  // it('test ngOnChanges()', () => {
+  //   component.ngOnChanges(changes);
+  //   fixture.detectChanges();
+  //   expect(component.testResultObj).toEqual({} as TestRecordTestType);
+  // });
+
+  it('should switch current state to view', () => {
```
alleypa, line 88: this is not needed. The converse below is sufficient

src/app/test-record/test-record.component.spec.ts
```diff
@@ -1,55 +1,128 @@
...
+    component.testStations = [{} as TestStation];
     fixture.detectChanges();
   });
 
   it('should create', () => {
     expect(component).toBeTruthy();
+    expect(fixture).toMatchSnapshot();
+  });
+
+  it('test ngOnInit()', () => {
```
alleypa, line 76: we can remove this test. component.onFormChanges is covered below

src/app/shared/directives/digit-only.directive.ts
irina005: tslint wants me to put app instead of vtm

src/app/test-record/defects/defects.component.html
irina005: as it's not in the scope of this ticket better do it when solving bugs from 10283 (test record view ticket) so it can be tested properly

src/app/store/selectors/ReferenceData.selectors.spec.ts
```diff
@@ -0,0 +1,32 @@
+import { selectFeature } from '@app/store/selectors/ReferenceData.selectors';
+import { Preparer } from '@app/models/preparer';
+import { TestStation } from '@app/models/test-station';
+import { getPreparers, getTestStations } from '@app/store/selectors/ReferenceData.selectors';
+
+describe('vehicleTechRecordModel selectors', () => {
+  const createState = ({
+    referenceData = {
+      preparers: {} as Preparer[],
```
alleypa, line 9: show some intended key values here. The reader will understand what preparers and  testStations are

src/app/store/reducers/VehicleTestResultModel.reducers.spec.ts
```diff
@@ -53,5 +57,52 @@ describe('VehicleTestResultModel Reducer', () => {
...
+
+      expect(result).toMatchObject({
+        ...initialVehicleTestResultModelState,
+        editState: VIEW_STATE.VIEW_ONLY
+      });
+    });
+  });
+
+  describe('[Update Test Result]', () => {
+    test('should update testStations', () => {
```
alleypa, line 73: adopt a consistent approach

src/app/technical-record/test-history/test-history.component.ts
```diff
@@ -8,28 +9,23 @@ import {TechRecordHelpersService} from '@app/technical-record/tech-record-helper
   changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class TestHistoryComponent implements OnInit {
+  @Input() testResultJson: TestResultModel[];
```
alleypa, line 12: use testResults please

src/app/technical-record/test-history/test-history.component.ts
```diff
@@ -8,28 +9,23 @@ import {TechRecordHelpersService} from '@app/technical-record/tech-record-helper
...
 
-  @Input() testResultJson: any;
+  constructor(public techRecHelpers: TechRecordHelpersService) {}
```
alleypa, line 14: we should remove this service when we can `TechRecordHelpersService`. not much significant

src/app/test-record/test-history/test-history.component.scss
```diff
@@ -0,0 +1,7 @@
+.t-history {
+  margin-bottom: 0 !important;
+}
+
+tr:last-child td {
```
alleypa, line 5: doing css like this will cause specificity hard to debug.

JonLazariniDD:
Why have we got 2 tickets (2 commits) in this PR?

src/app/models/vehicle-test-result-update.ts
```diff
@@ -0,0 +1,7 @@
+import { UserDetails } from '@app/models/user-details';
+import { TestResultModel } from '@app/models/test-result.model';
+
+export interface VehicleTestResultUpdate {
+  testResult: TestResultModel;
+  msUserDetails: UserDetails;
```
JonLazariniDD, line 6: Why have we got the msUserDetails in there? If i am not wrong, this is not part of the Techrecord or testResult object.
If we need it in the view it should probably be at the root of the store in a separate reducer so that it can be shared easily
```
{
  techRecord: [],
  loading: false,
  error: null,
  msUser: null,
  ...
}
```
it is worth having a look at the API spec with the swagger file.
https://github.com/dvsa/cvs-svc-technical-records/blob/develop/tests/resources/technical-records.json
https://github.com/dvsa/cvs-svc-test-results/blob/develop/tests/resources/test-results.json

Also why do we need a separate model to update the test result?
Can we just have a vehicle-test-result model?

src/app/shared/components/autocomplete/autocomplete.component.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+    expect(component.autocompleteInput.value).toBe('test write');
+  });
+
+  it('should change value on registerOnChange', () => {
+    component.registerOnChange('');
+    expect(component.autocompleteInput.value).toBe(null);
+  });
+
+  it('should change value on registerOnTouched', () => {
+    component.registerOnTouched('');
```
JonLazariniDD, line 68: can we add an expectation that checks when the state of the component is changing like if a user interaction was happening please?

src/app/shared/components/autocomplete/autocomplete.component.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  @Input()
+  set id(value: string) {
+    this.internalId = value;
+    this.externalId = null;
+  }
+  get id() {
+    return `${this.internalId}`;
+  }
+  private internalId = '';
```
JonLazariniDD, line 68: can we move all the initialisation or definition at the top of the file so it is consistent and we know what to expect as a behaviour from the component please? 
For example, private variables, ngForm, @Input, @Output, getters, setters, constructor, methods.
It will make things easier to reason about if we can identify them straight away

src/app/shared/components/autocomplete/autocomplete.component.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+    expect(testFixture).toMatchSnapshot();
+  }));
+
+  it('should correctly write value into input field', () => {
+    component.writeValue('test write');
+    expect(component.autocompleteInput.value).toBe('test write');
+  });
+
+  it('should change value on registerOnChange', () => {
+    component.registerOnChange('');
```
JonLazariniDD, line 63: I think we also have to spy onchange and check that the cb is called.
Also can test the behaviour of the component on the dom please?
What will happen if onChange has a value?

src/app/shared/components/autocomplete/autocomplete.component.ts
JonLazariniDD, line 68: Can we please try to group all the things that belong together at the same place and same groups, it will make it easier to reason about the component when we read at the definition.
We could have something like @input, @output, private vars, var, getters, setters, all the angular things (lifecyclem methods).

src/app/shared/components/autocomplete/autocomplete.component.ts
```diff
@@ -0,0 +1,103 @@
...
+  set value(val) {
+    this._value = val;
+    this.onchange(val);
+  }
+  get value() {
+    return this._value;
+  }
+
+  @HostBinding('attr.id')
+  externalId = '';
```
JonLazariniDD, line 58: We have to have the value given from the database if not we might want to look at uuid, it is worth checking with the BE if there are unique value we could be using

src/app/shared/directives/digit-only.directive.spec.ts
```diff
@@ -0,0 +1,18 @@
...
+
+describe('DigitOnlyDirective', () => {
+  it('should create an instance', () => {
+    const directive = new DigitOnlyDirective();
+    expect(directive).toBeTruthy();
+  });
+  it('should call prevent default on non numeric key press', () => {
+    const directive = new DigitOnlyDirective();
+    const event: KeyboardEvent = {
+      key: 'a',
```
JonLazariniDD, line 11: can we also add ' ' so we get both conditions and branching covered please?

src/app/store/actions/ReferenceData.actions.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+  it('action should have correct type & payload', () => {
+    const actionInstance = new LoadPreparersSuccess({} as Preparer[]);
+    expect(actionInstance.type).toBe(EReferenceDataActions.LoadPreparersSuccess);
+  });
+});
+
+describe('LoadTestStationsSuccess', () => {
+  it('action should have correct type & payload', () => {
+    const actionInstance = new LoadTestStationsSuccess({} as TestStation[]);
+    expect(actionInstance.type).toBe(EReferenceDataActions.LoadTestStationsSuccess);
```
JonLazariniDD, line 35: Would it be possible to assert the initial payload it returns when instantiated?

src/app/store/actions/VehicleTestResultModel.actions.spec.ts
```diff
@@ -1,60 +1,88 @@
...
-  vehicleSize: '213',
-  preparerName: '213'
-}
+  vehicleSize: '213'
+} as TestResultModel;
 
 describe('GetVehicleTestResultModel', () => {
-  test('action should have the right type and payload when used', () => {
-    const actionInstance = new GetVehicleTestResultModel(testResultModelExample);
+  it('action should have the right type and payload when used', () => {
```
JonLazariniDD, line 23: not sure we need 'action' in the unit test description.

src/app/store/actions/VehicleTestResultModel.actions.spec.ts
```diff
@@ -1,60 +1,88 @@
...
+      EVehicleTestResultModelActions.GetVehicleTestResultModelSuccess
+    );
+    expect(actionInstance.payload).toStrictEqual([testResultModelExample]);
   });
 });
 
 describe('GetVehicleTestResultModelFailure', () => {
-  test('action should have the right type and payload when used', () => {
-    const actionInstance = new GetVehicleTestResultModelFailure(testResultModelExample);
+  it('action should have the right type and payload when used', () => {
```
JonLazariniDD, line 43: same for 'action'

src/app/store/actions/VehicleTestResultModel.actions.spec.ts
```diff
@@ -1,60 +1,88 @@
...
+      EVehicleTestResultModelActions.GetVehicleTestResultModelFailure
+    );
+    expect(actionInstance.payload).toBe('test');
   });
-});
\ No newline at end of file
+});
+
+describe('SetCurrentState', () => {
+  it('action should have correct type & payload', () => {
```
JonLazariniDD, line 54: same for 'action'

src/app/store/actions/VehicleTestResultModel.actions.spec.ts
```diff
@@ -1,60 +1,88 @@
...
+    const actionInstance = new SetCurrentState(VIEW_STATE.VIEW_ONLY);
+
+    expect(actionInstance.type).toBe(EVehicleTestResultModelActions.SetCurrentState);
+    expect(actionInstance.editState).toBe(VIEW_STATE.VIEW_ONLY);
+  });
+});
+
+describe('UpdateTestResult', () => {
+  it('action should have correct type & payload', () => {
```
JonLazariniDD, line 63: same for 'action'

src/app/store/actions/VehicleTestResultModel.actions.spec.ts
```diff
@@ -1,60 +1,88 @@
...
+    const actionInstance = new UpdateTestResult({testResultUpdated: testResultModelExample, testTypeNumber: '1'});
+
+    expect(actionInstance.type).toBe(EVehicleTestResultModelActions.UpdateTestResult);
+    expect(actionInstance.testResultTestTypeNumber.testResultUpdated).toStrictEqual(testResultModelExample);
+    expect(actionInstance.testResultTestTypeNumber.testTypeNumber).toStrictEqual('1');
+  });
+});
+
+describe('SetSelectedTestResultModel', () => {
+  it('action should have correct type & payload', () => {
```
JonLazariniDD, line 73: same for 'action'

src/app/store/actions/VehicleTestResultModel.actions.spec.ts
```diff
@@ -1,60 +1,88 @@
...
+    const actionInstance = new SetSelectedTestResultModel('');
+
+    expect(actionInstance.type).toBe(EVehicleTestResultModelActions.SetSelectedTestResultModel);
+    expect(actionInstance.payload).toBe('');
+  });
+});
+
+describe('SetSelectedTestResultModelSuccess', () => {
+  it('action should have correct type & payload', () => {
```
JonLazariniDD, line 82: same for 'action'

src/app/store/effects/VehicleTestResultModel.effects.ts
```diff
@@ -1,32 +1,73 @@
...
+    switchMap((testResults: TestResultModel[]) => {
+      this._store.dispatch(new ClearErrorMessage());
+      return of(new GetVehicleTestResultModelSuccess(testResults));
+    })
+  );
+
+  @Effect()
+  updateTestResult$ = this._actions$.pipe(
+    ofType<UpdateTestResult>(EVehicleTestResultModelActions.UpdateTestResult),
+    map((action) => action.testResultTestTypeNumber),
```
JonLazariniDD, line 41: can simplify this by destructuring it I think 
```
map(({testResultTestTypeNumber}) => testResultTestTypeNumber)
```

src/app/store/effects/VehicleTestResultModel.effects.ts
```diff
@@ -1,32 +1,73 @@
...
+    switchMap((testResultTestTypeNumber: TestResultTestTypeNumber) => {
+
+      const testResultTestTypeNumberObject = testResultTestTypeNumber;
```
JonLazariniDD, line 44: It seems that we need certain props from `testResultTestTypeNumber ` here to reconstruct a new object
Can we do something like to make it simple instead rather than creating this temporary object?
```
// pulling out testResultUpdated from testResultTestTypeNumber
const { testResultUpdated } = testResultTestTypeNumber
...
testResultObject.testResult = testResultUpdated
```

src/app/store/reducers/VehicleTestResultModel.reducers.spec.ts
JonLazariniDD: use it agree and also needs to be more descriptive please like mentioned above. There is more happening to it.
this will be more legible and intelligible.

src/app/store/reducers/VehicleTestResultModel.reducers.spec.ts
```diff
@@ -53,5 +57,53 @@ describe('VehicleTestResultModel Reducer', () => {
...
+      } as TestResultTestTypeNumber;
 
+      initialVehicleTestResultModelState.vehicleTestResultModel = [
+        { testTypes: [{ testNumber: '1' }] } as TestResultModel,
+        { testTypes: [{ testNumber: '2' }] } as TestResultModel
+      ];
+
+      const action = new UpdateTestResult(vehicleTestResultTestTypeNumber);
+
+      const updatedVehicleTestResults = initialVehicleTestResultModelState.vehicleTestResultModel.map(
```
JonLazariniDD, line 86: I will need to have an understanding of this please.
I am confused, it is not showing a simple
 arrange act assert pattern
We already defined the initial state shape a few lines above

src/app/store/reducers/VehicleTestResultModel.reducers.ts
```diff
@@ -1,20 +1,61 @@
...
+  action: VehicleTestResultModelActions
+) {
   switch (action.type) {
     case EVehicleTestResultModelActions.GetVehicleTestResultModelSuccess: {
       return {
         ...state,
-        selectedTestResultModel: action.payload,
-        error: null // clear error message
+        vehicleTestResultModel: action.payload,
+        error: null
```
JonLazariniDD, line 16: what is this error object? Shouldn't be part of the root reducer instead.
This will need to be moved

src/app/store/reducers/VehicleTestResultModel.reducers.ts
```diff
@@ -1,20 +1,61 @@
...
+    case EVehicleTestResultModelActions.SetCurrentState: {
+      return {
+        ...state,
+        editState: action.editState
+      };
+    }
+
+    case EVehicleTestResultModelActions.UpdateTestResultSuccess: {
+      const updatedVehicleTestResults = state.vehicleTestResultModel.map((testResult) => {
+        if (
```
JonLazariniDD, line 36: this can be moved into a state selector I believe,
I am not convinced storing logic in a reducer in the appropriate way.
@alleypa @m-ujaffer what's your view on that please?
I would tend to try to have composition as much possible with reducers simply rendering data, any logic should happen in selectors that we could compose to reshape the state the way we want it This will extract out as much logic as possible which could be handled in some utils functions which will be way easier to test than a possible bloated reducer.

src/app/store/selectors/VehicleTestResultModel.selectors.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+  });
+
+  it('selectTestTypeById', () => {
+    const state = createState();
+    const testRes = state.vehicleTestResultModel.vehicleTestResultModel.vehicleTestResultModel;
+    select(selectTestTypeById('1'));
+    expect(Object.keys(testRes)).toHaveLength(2);
+    Object.keys(testRes).forEach((res) => {
+      expect(testRes[res].testTypes).toHaveLength(1);
+      // spyOn(testRes[res].testTypes, 'some');
```
JonLazariniDD, line 46: commented out

src/app/technical-record-search/test-result.service.spec.ts
```diff
@@ -74,6 +75,18 @@ describe('TestResultService', () => {
     req.flush({ mockObject: 'mock' });
   });
 
+  it('updateTestResults should update', (done) => {
+    const mock = { mockObject: 'mock' };
+    service.updateTestResults('123', {} as VehicleTestResultUpdate).subscribe((res) => {
+      expect(res).toBeDefined();
```
JonLazariniDD, line 81: Can we clear the mock after each test?
Can we spy on the response to check that it is sending the correct body  '123' or have the correct response?

src/app/technical-record/technical-record.component.ts
```diff
@@ -1,116 +1,116 @@
...
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class TechnicalRecordComponent implements OnInit {
+  showAdrDetails: boolean;
+  adrDisplayParams: { [key: string]: boolean };
+  techRecord: FormGroup;
+
+  allOpened = false;
+  panels: { panel: string; isOpened: boolean }[] = [
+    { panel: 'panel1', isOpened: false },
```
JonLazariniDD, line 35: we need more descriptive names than panel1, etc...

src/app/technical-record/technical-record.component.ts
```diff
@@ -1,116 +1,116 @@
...
+      width: '600px',
+      data: {
+        context: 'Enter reason for changing technical record'
+      }
+    });
+
+    dialogRef.afterClosed().subscribe((result) => {
+      if (result && result.isSave) {
+        mergedRecord.reasonForCreation = result.data;
+        this.submitTechRecord.emit(mergedRecord);
```
JonLazariniDD, line 111: I think it would make things easier if we use ngrx and showed in in the state instead? What do you think?

src/app/technical-record/technical-record.component.ts
```diff
@@ -1,116 +1,116 @@
...
+      }
+    });
+    // }
+  }
```
JonLazariniDD, line 115: Have we got unit tests for this file?

src/app/technical-record/test-history/test-history.component.html
```diff
@@ -1,54 +1,77 @@
...
-  <tr class="govuk-table__row">
-    <th scope="col" class="govuk-table__header">Test type</th>
-    <th scope="col" class="govuk-table__header">Test date</th>
-    <th scope="col" class="govuk-table__header">Expiry date</th>
-    <th scope="col" class="govuk-table__header">Certificate number</th>
-    <th scope="col" class="govuk-table__header">Result</th>
-    <th scope="col" class="govuk-table__header">Tester name</th>
-    <th scope="col" class="govuk-table__header"> </th>
-  </tr>
+    <tr class="govuk-table__row">
```
JonLazariniDD, line 3: indentation is off I think

src/app/technical-record/test-history/test-history.component.html
```diff
@@ -1,54 +1,77 @@
...
+    >
+      <tr
+        class="govuk-table__row"
+        *ngFor="
+          let testType of testRes.testTypes | SortByDatePipe: 'testTypeEndTimestamp';
+          let ii = index
+        "
+      >
+        <td id="{{ 'test-testCode-tRes-' + i + '-tType-' + ii }}" class="govuk-table__cell">
+          {{ techRecHelpers.capitalizeWord(testType.testCode | DefaultNullOrEmpty) }}
```
JonLazariniDD, line 25: Can we move this into the component itself please, it will make harder to test logic when it is inlined.

src/app/technical-record/test-history/test-history.component.html
```diff
@@ -1,54 +1,77 @@
...
-      <td>-</td>
-      <td>-</td>
-      <td>-</td>
-      <td>-</td>
-      <td>-</td>
-    </tr>
+      <tr>
+        <td>-</td>
+        <td>-</td>
+        <td>-</td>
```
JonLazariniDD, line 69: i think the indentation is off

src/app/technical-record/test-history/test-history.component.ts
```diff
@@ -8,28 +9,23 @@ import {TechRecordHelpersService} from '@app/technical-record/tech-record-helper
...
 
-
-  constructor(public techRecHelpers: TechRecordHelpersService) {
-  }
-
-  ngOnInit() {
-  }
+  ngOnInit() {}
 
   setResultColor(testResult: string) {
```
JonLazariniDD, line 18: Can we use scss classes instead of doing it with js.
It will be easier to maintain and also in case the colours change, the snapshot will tell us and this will help for regression

src/app/test-record/defects/defects.component.html
```diff
@@ -0,0 +1,92 @@
...
+      {{ defect.deficiencyCategory }}
+    </span>
+  </h2>
+  <div class="govuk-summary-list__row">
+    <dt class="govuk-summary-list__key">
+      Category
+    </dt>
+    <dd class="govuk-summary-list__value" id="test-imNumber-imDescription">
+      {{
+        concatStrings([defect.imNumber, defect.imDescription], ' ')
```
JonLazariniDD, line 20: can we use a pipe instead of having that inlined?

src/app/test-record/defects/defects.component.scss
```diff
@@ -0,0 +1,25 @@
+
+.govuk-summary-list {
+
+  .tag-custom {
+    padding: 1px 10px 1px 10px;
+    vertical-align: middle;
+  }
+
+  & .tag--black {
+    background-color: #0b0c0c !important;
```
JonLazariniDD, line 10: Same comment as the one made for the scss earlier

src/app/test-record/defects/defects.component.scss
```diff
@@ -0,0 +1,25 @@
+
+.govuk-summary-list {
+
+  .tag-custom {
+    padding: 1px 10px 1px 10px;
```
JonLazariniDD, line 5: could be short handed with 1px 10px also if this a custom component this will need to ported into the global scss file at the bottom.
We could follow the convention govuk lib uses
// colours

// typography

...

/**
** Components
**/

// tags
your scss tag

Also let's avoid using @extend, it is bad practice in general,
this scopes to the parent component and increase specificity making things very complicated if we want to override styles later or use composition (with scss conventions like BEM for example).
It is a great tool when architecting the scss to make it dry but will require some thorough thinking beforehand.

src/app/test-record/defects/defects.component.scss
```diff
@@ -0,0 +1,25 @@
...
+    vertical-align: middle;
+  }
+
+  & .tag--black {
+    background-color: #0b0c0c !important;
+    @extend .tag-custom;
+  }
+
+  & .tag--red {
+    background-color: #d4351c !important;
```
JonLazariniDD, line 15: Same comment as the one made for the scss earlier

src/app/test-record/defects/defects.component.scss
```diff
@@ -0,0 +1,25 @@
...
+    @extend .tag-custom;
+  }
+
+  & .govuk-summary-list__row:last-child {
+    & .govuk-summary-list__value, .govuk-summary-list__key {
+      border-bottom: none !important;
```
JonLazariniDD, line 21: Same comment as the one made for the scss earlier

src/app/test-record/defects/defects.component.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+    expect(fixture).toMatchSnapshot();
+  });
+
+  it('should concat strings', () => {
+    const stringsConcatenated = component.concatStrings(['test1', 'test2'], '/');
```
JonLazariniDD, line 34: This is probably a pipe instead or a util function taht can be shared. it is probably best to have in such file than bloat the component with that.

src/app/test-record/defects/defects.component.ts
```diff
@@ -0,0 +1,23 @@
...
+})
+export class DefectsComponent implements OnInit {
+  @Input() testType: TestType;
+  @Input() testRecord: TestResultModel;
+
+  constructor() {}
+
+  ngOnInit() {}
+
+  concatStrings(strings, concatBy): string {
```
JonLazariniDD, line 19: probably best use a pipe instead

src/app/test-record/emission-details/emission-details.component.scss
```diff
@@ -0,0 +1,7 @@
+.govuk-summary-list {
+  & .govuk-summary-list__row:last-child {
+    & .govuk-summary-list__value, .govuk-summary-list__key {
+      border-bottom: none !important;
```
JonLazariniDD, line 4: same for scss earlier.

src/app/test-record/guards/preparers.guard.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+    spyOn(store, 'dispatch').and.callThrough();
+  }));
+
+  describe('canActivate', () => {
+    it('should return true for preparers in state', () => {
+      testResultService = { getPreparers: () => true };
+      preparersGuard = new PreparersGuard(store, testResultService);
+      expect(preparersGuard.populateStoreWithDataFromApi).toBeTruthy();
+      expect(preparersGuard.hasPreparers).toBeTruthy();
+      expect(preparersGuard.canActivate()).toBeDefined();
```
JonLazariniDD, line 39: Could we also assert the store to see that it is populated with the correct data?

src/app/test-record/guards/test-stations.guard.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+      ]
+    }).compileComponents();
+    store = TestBed.get(Store);
+    spyOn(store, 'dispatch').and.callThrough();
+  }));
+
+  describe('canActivate', () => {
+    it('should return true for preparers in state', () => {
+      testResultService = { getTestStations: () => true };
+      testStationsGuard = new TestStationsGuard(store, testResultService);
```
JonLazariniDD, line 36: can we inspect the store to see it has the correct data?

src/app/test-record/notes/notes.component.scss
```diff
@@ -0,0 +1,7 @@
+.govuk-summary-list {
+  & .govuk-summary-list__row:last-child {
+    & .govuk-summary-list__value, .govuk-summary-list__key {
+      border-bottom: none !important;
```
JonLazariniDD, line 4: same for scss

src/app/test-record/test-record.mapper.ts
```diff
@@ -0,0 +1,288 @@
...
+        return TEST_STATION_TYPE.GVTS;
+        break;
+      case 'hq':
+        return TEST_STATION_TYPE.HQ;
+        break;
+    }
+  }
+
+  getTestTypeApplicable(sectionName: string) {
+    const testTypesApplicable = {
```
JonLazariniDD, line 33: Can this be stored in a config file with enums instead?
I believe this could be reused for unit tests andpossible other vehicle types?

src/app/test-record/test-record.mapper.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+  });
+
+  it('should map form values for PSV', () => {
+    const testResultObject = {
+      testRecord: TESTING_TEST_MODELS_UTILS.mockTestRecord(),
+      testType: TESTING_TEST_MODELS_UTILS.mockTestType()
+    } as TestRecordTestType;
+
+    const testResultMapped = mapper.mapFormValues(
+      { testType: { testTypeId: '1', seatbeltInstallationCheckDate: true } },
```
JonLazariniDD, line 61: Can we extend the test in a way that it shows everything returned instead please?

src/app/test-record/test-record.mapper.ts
```diff
@@ -0,0 +1,288 @@
...
+    testTypeMapped.smokeTestKLimitApplied = testResultFormData.testType.smokeTestKLimitApplied;
+    testTypeMapped.fuelType = testResultFormData.testType.fuelType;
+    testTypeMapped.modType = testResultFormData.testType.modType;
+    testTypeMapped.modificationTypeUsed = testResultFormData.testType.modificationTypeUsed;
+    testTypeMapped.particulateTrapFitted = testResultFormData.testType.particulateTrapFitted;
+    testTypeMapped.particulateTrapSerialNumber =
+      testResultFormData.testType.particulateTrapSerialNumber;
+    testTypeMapped.additionalNotesRecorded = testResultFormData.testType.additionalNotesRecorded;
+
+    if (
```
JonLazariniDD, line 217: I think we will need to extend the units for this function

src/app/test-record/test-section/test-section.component.scss
```diff
@@ -0,0 +1,7 @@
+.govuk-summary-list {
+  & .govuk-summary-list__row:last-child {
+    & .govuk-summary-list__value, .govuk-summary-list__key {
+      border-bottom: none !important;
```
JonLazariniDD, line 4: same for scss

src/app/test-record/vehicle/vehicle.component.scss
```diff
@@ -0,0 +1,7 @@
+.govuk-summary-list {
+  & .govuk-summary-list__row:last-child {
+    & .govuk-summary-list__value, .govuk-summary-list__key {
+      border-bottom: none !important;
```
JonLazariniDD, line 4: same for scss

src/app/test-record/visit/visit-edit/visit-edit.component.scss
```diff
@@ -0,0 +1,10 @@
+#test-testStationType {
+  border: none;
+}
+#test-testStationType:focus {
+  outline: none;
+  box-shadow: none;
+}
+#test-testStationType:disabled {
+  color: #0b0c0c;
```
JonLazariniDD, line 9: same for scss
Also i would be very careful to use and id for it, a class should be enough, we don't need to increase specificity for nothing,
We could probably port those variables over to a separate scss file if we can't use govuk but i suspect we should be able to with the appropriate markup in place

src/app/test-record/visit/visit.component.scss
```diff
@@ -0,0 +1,7 @@
+.govuk-summary-list {
+  & .govuk-summary-list__row:last-child {
+    & .govuk-summary-list__value, .govuk-summary-list__key {
+      border-bottom: none !important;
```
JonLazariniDD, line 4: same

src/assets/config/config.deploy.json
```diff
@@ -27,6 +27,7 @@
     "APITechnicalRecordServerUri": "https://#{apiServerPrefix}.demo.com/api/v1.0/",
     "APITestResultServerUri": "https://#{apiServerPrefix}.demo.com/api/v1.0/",
     "APIDocumentsServerUri": "https://api.nonprod.cvs.dvsacloud.uk/BRANCH",
-    "APIDocumentBlobUri": "https://api.nonprod.cvs.dvsacloud.uk/BRANCH/cvs-nonprod-adr-pdfs/BRANCH%2F"
+    "APIPreparersServerUri": "https://api.nonprod.cvs.dvsacloud.uk/BRANCH",
```
JonLazariniDD, line 30: please remove this file

src/assets/config/config.dev.json
```diff
@@ -1,7 +1,6 @@
-{
-  "env": {
-    "name": "dev"
-  },
+{  "env": {
+  "name": "dev"
+},
```
JonLazariniDD, line 3: and this file

src/app/models/test-result.model.ts
```diff
@@ -1,19 +1,21 @@
...
 
 export interface TestResultModel {
+  reasonForCreation: string;
+  testResultId: string;
   testerStaffId: string;
-  testStartTimestamp: Date;
+  testStartTimestamp: string;
   odometerReadingUnits: string;
-  testEndTimestamp: Date;
+  testEndTimestamp: string;
```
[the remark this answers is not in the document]
alleypa: Hi @JonLazariniDD, I suggested to use the string type for dates partly because 
- the API specs specifies them as string format
- it will be easy to parse and implement
- but the onus is on the dev to ensure the string format is returned to the back end as expected i.e `YYYY-MM-DD`, `YYYY-MM-DD hh:mm:ss`

src/app/models/vehicle-test-result-update.ts
alleypa: I think the problem here is back end expects a different object to allow savings. We have a singleton that will give us the user details as expected

src/app/store/reducers/VehicleTestResultModel.reducers.ts
alleypa: Yes I agree @JonLazariniDD. Reducers are not meant to be complicated. I think an essential selector could be missing or the need extra functions out

src/app/technical-record/technical-record.component.ts
alleypa: Ideally this file should not appear in this ticket. but due to code change in `line 51`. I agree with the panel naming. The code is a tech debt. We have a new  component for implementing accordion. So we can definitely refactor this in the another ticket

src/app/technical-record/technical-record.component.ts
alleypa: Yes I agree this can be solved in NGRX way. We would need to extract this implementation and the dialog component implementation and solve in NGRX way. Not sure on when or how. I will discuss with you @JonLazariniDD

src/app/technical-record/technical-record.component.ts
alleypa: This will be picked up as part of 10202.

src/app/test-record/seatbelt-installation-check/seatbelt-installation-check-edit/seatbelt-installation-check-edit.component.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+  it('should set seatbeltOptionSelected to Yes', () => {
+    component.testType = TESTING_TEST_MODELS_UTILS.mockTestType({
+      seatbeltInstallationCheckDate: true
+    });
+    component.ngOnInit();
+    expect(component.seatbeltOptionSelected).toEqual('Yes');
+  });
+
+  it('should set seatbeltOptionSelected to No', () => {
+    component.testType = TESTING_TEST_MODELS_UTILS.mockTestType({
```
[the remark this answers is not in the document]
alleypa: please rename as TESTING_TESTMODELS_UTILS

src/app/models/vehicle-test-result-update.ts
JonLazariniDD: I don't remember seeing that as needed on the POST request and this won't be required I believe: https://github.com/dvsa/cvs-svc-test-results/blob/develop/tests/resources/test-results-post.json

src/app/technical-record/technical-record.component.ts
JonLazariniDD: let's create a subtask for it please if we decide to do it later, the code coverage is good at the moment but we have a reminder and put a reference to it here please.
I think it would be best to have them pushed as commit code now but if it will take time or need more clarity we should reference it here and create a task.
Also bear in mind that tech-record will have different fields in fact, that will not be a 'copy paste' job.

irina005:
Closing as per the decision that validation tickets will not be prioritized

PR #130, opened 2020-04-24
Feature/cvsb 10202 HGV creation 

rspasici:


alleypa:
dropping so as to point to common branch

PR #139, opened 2020-05-05
feature/CVSB-14720 - [BUG] Test record history is not displayed properly when going to tech record view page from tech record selection page

irina005:
When a user goes from tech record selection screen to tech record view page, the BE request on test-results to retrieve all test records for that vehicle is no longer triggered, which means that the test record history will always be empty, even if the vehicle actually has test records on it.

https://jira.dvsacloud.uk/browse/CVSB-14720


src/app/store/effects/VehicleTechRecordModel.effects.ts
```diff
@@ -74,7 +73,7 @@ export class VehicleTechRecordModelEffects {
     ),
     switchMap((action) => {
       this.router.navigate(['/technical-record']);
-      return [new SetSelectedVehicleTechnicalRecordSucess(action.vehicleTechRecord)];
+      return [new GetVehicleTestResultModel(action.vehicleTechRecord.systemNumber), new SetSelectedVehicleTechnicalRecordSucess(action.vehicleTechRecord)];
```
alleypa, line 76: return [
  new GetVehicleTestResultModel(action.vehicleTechRecord.systemNumber), 
  new SetSelectedVehicleTechnicalRecordSucess(action.vehicleTechRecord)
];

irina005:
Closing in order to raise a PR having **common** as base branch

PR #140, opened 2020-05-06
DEVELOP bug fixing

irina005:
- feature/CVSB-14720 - https://jira.dvsacloud.uk/browse/CVSB-14720
- feature/CVSB-12898 - https://jira.dvsacloud.uk/browse/CVSB-12898


src/app/store/reducers/error.reducers.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  describe('when initialised', () => {
+    beforeEach(() => {
+      action = {} as Action;
+      const initState: ErrorState = { errors: [] };
+      resultState = ErrorReducer(initState, action);
+    });
+
+    it('should set the default state', () => {
+      expect(resultState).toMatchSnapshot();
```
JonLazariniDD, line 19: Do we want to have an expected state to assert on instead of using the snapshot to make sure we don't have false positive in the future?
Let's say if we agree that the app state is
```
error: {},
loading,
techRecord,
...
```
if we decide to change the error state to `error: {list: []}` the test will pass and didn't check whether we inadvertently modified the shape of the state or not.
I am not very familiar with snapShot approach on reducers so i would like to hear your thoughts on that please.
I would tend to define the expectedState shape like this instead of using snapShot to follow the "arrange, act, assert" pattern
```
// example for setting the error in the reducer
const action = {type: SetErrorMessage, payload: ['new error']}
const expectedState = {errors: ['new error'] }

expect(ErrorReducer(action, initialErrorState)).toMatch(expectedState)
```
With this approach above you can clearly read the effect (expectedState) it will have when the reducer receives the action. If we decide to change the state shape it will fail and it is good that we know so we can change it where it is necessary in the codebase.

src/app/store/reducers/error.reducers.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+
+describe('ErrorReducer', () => {
+  let state: ErrorState;
+  let resultState: ErrorState;
+  let action: any;
+
+  describe('when initialised', () => {
+    beforeEach(() => {
+      action = {} as Action;
+      const initState: ErrorState = { errors: [] };
```
JonLazariniDD, line 14: we might also use initialErrorState for this initialErrorState.
Also, would want to create an InitialState Object that we could reference in the unit tests files so we don't hardcode the value in the test and this will act as the single source of truth?
Or have an object with a default state that matches the IAppState shape?

I think we already have an app.state.ts, could use that instead?
```
{
  router: ['/'],
  loader: false,
  errors: null,
  vehicleTechRecordModel: [],
  vehicleTestRecordModel: []
}

// in the test
import {initialState} from 'default-state'

// use the state slices to set the tests and reducers
initialState.errors
...

```

src/app/store/reducers/error.reducers.ts
```diff
@@ -0,0 +1,33 @@
...
+): ErrorState {
+  switch (action.type) {
+    case EErrorActions.SetErrorMessage: {
+      return {
+        ...state,
+        errors: [...action.payload]
+      };
+    }
+
+    case EErrorActions.ClearErrorMessage: {
```
JonLazariniDD, line 23: shouldn't be null instead? i think we define the default value as null which is probably when we have no error?
Also, shouldn't initialErrorState be null instead of [] in fact?
I suspect that with the empty array we could still map over it hence the errorComponent could show with empty messages?

src/app/technical-record-create/technical-record-create.component.scss
```diff
@@ -1,22 +1,4 @@
...
-  }
-
-  & .govuk-error-summary {
-    padding-bottom: 0 !important;
-    & .govuk-error-message {
-      text-decoration: underline;
-    }
-  }
+.govuk-back-link {
+  text-decoration: underline;
```
JonLazariniDD, line 2: we will have to check on that one as i suspect we are modifying the style of the back button to match some design or based on feedback however by doing so we are not being compliant with GDS design system.
I am happy to pick that up separately in fact if we need @irina005 as I believe we should be using GDS living styleguide out of the box and if changes are required then the team should have had a request submitted to check whether they agree on it.

src/app/shared/error-summary/error-summary.component.html
```diff
@@ -1,5 +1,5 @@
 <div *ngIf="errorList$ | async as errorList" class="govuk-width-container">
-  <main class="govuk-main-wrapper" id="errorBox">
+  <main class="govuk-main-wrapper" id="errorBox" *ngIf="errorList.length">
```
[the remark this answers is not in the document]
alleypa: yes I agree we should use `id="errorBox"`

src/app/shared/error-summary/error-summary.component.spec.ts
```diff
@@ -1,40 +1,68 @@
...
+import { BehaviorSubject } from 'rxjs';
+import { map } from 'rxjs/operators';
+
 import { ErrorSummaryComponent } from './error-summary.component';
-import { hot } from 'jasmine-marbles';
+import { getErrors } from '@app/store/selectors/error.selectors';
+
+const mockSelector = new BehaviorSubject<any>(undefined);
+
+class MockStore {
```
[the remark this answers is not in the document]
alleypa: yes I agree we should separate. This pattern has just been introduced @alleypa will take care of this

src/app/store/reducers/error.reducers.spec.ts
alleypa: I agree. We shoud replace line 14 with `const initState = initialAppState.error;`

src/app/store/reducers/error.reducers.spec.ts
alleypa: The generate snapshot should be the guide here and simplifies testing of reducers. So if any changes in shape occurs the test will fail

src/app/store/reducers/error.reducers.ts
alleypa: This approach is preferred greatly because it gives better API flexibility in terms of checking for the existence of errors in the state. It also matches the expected response received from back end which are string array of errors

JonLazariniDD: Thanks for the clarification.
Shouldn't we have this instead (of null for the default state) if it is the case?
```
export interface ErrorState {
  errors: string[] | [];
}
```

src/app/store/reducers/error.reducers.spec.ts
JonLazariniDD: thanks, ok so we are saying we will localise defaultState and state interfaces in the reducer I believe. it is also a great approach if we wish to export them as separate module (like the redux ducks pattern) they will also have to be mapped/imported in IAppState and relevant unit test files for consistency and maintainability.
At the moment IAppState has the following if I am not wrong

```
 error?: [string] | null; // i guess this should be replaced by ErrorState as well?
```

src/global.scss
```diff
@@ -0,0 +1,52 @@
...
+// Spacing
+
+// Typography
+
+
+/**************/
+// COMPONENTS //
+/**************/
+
+.govuk-back-link {
```
JonLazariniDD, line 37: As agreed, a tech improvement ticket will be created to address the GDS styling we created here,
this is essentially porting the existing style into this component for override which shouldn't be necessary.

Reference: https://jira.dvsacloud.uk/browse/CVSB-15034

PR #141, opened 2020-05-07
Feature/cvsb 10175 detailed psv view

rspasici:


src/app/technical-record/body/body.component.html
```diff
@@ -1,5 +1,5 @@
 <dl class="technical-record-sub-details govuk-summary-list ">
-  <div class="govuk-summary-list__row">
+  <div class="govuk-summary-list__row" *ngIf="activeRecord.vehicleType === 'hgv' || activeRecord.vehicleType === 'trl'">
```
JonLazariniDD, line 2: I think we need to tidy up the section so we reduce the size of the file for this markup as it is going to grow exponentially and could end up having easily thousand lines of code due to the nature of the form please.

we will need to have a component that shows the common form fields elements across vehicle types.
In this instance,  body component has the following as common form field elements: Make, Model, body type, function, conversion reference.
In this template we could have them as the top as they are common:
```
// pseudo code
<ng-container *ngIf="!edit">
  ...
  <Make/>
  <Model/>
  <Bodytype/>
  <function/> 
  <conversion-reference/>
  // this contains specific fields for hgv 
  <vtm-vehicle-hgv *ngIf="activeRecord.vehicleType === 'hgv'"></vtm-vehicle-hgv>
  // this contains specific fields for trl
  <vtm-vehicle-trl *ngIf="activeRecord.vehicleType === 'trl'"></vtm-vehicle-trl>
  // this contains specific fields for psv
  <vtm-vehicle-psv *ngIf="activeRecord.vehicleType === 'psv'"></vtm-vehicle-psv>
</ng-container>
<ng-container *ngIf="edit">
  <vtm-*-edit [props]="props"></vtm-*-edit [props]="props>
</ng-container>
```
I think @m-ujaffer & @alleypa are taking a similar approach and were keen on using that.
We can decide later whether we move every form field elements (make, etc..) into their own separate components.

PR #171, opened 2020-05-26
Feature/CVSB-12375: edit test result - test section

irina005:
Ticket:
https://jira.dvsacloud.uk/browse/CVSB-12375


src/app/shared/components/autocomplete/autocomplete.component.scss
```diff
@@ -0,0 +1,3 @@
+input::-webkit-calendar-picker-indicator {
```
JonLazariniDD, line 1: i would assume this style will be used for every auto complete components so it is better to have in the global.scss assuming the styling is not given by the design system we have?
This would be some portable style should we had to change js framework.

src/app/shared/dialog-box-confirmation/dialog-box-confirmation.component.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+
+  describe('close', () => {
+    it('should close the modal when called', () => {
```
JonLazariniDD, line 35: need one more test for save

src/app/shared/dialog-box/dialog-box.component.scss
```diff
@@ -0,0 +1,18 @@
+textarea {
+  width: 80% !important;
+}
+
+.cancel {
```
JonLazariniDD, line 5: Can use the secondary button style for it

src/app/shared/dialog-box/dialog-box.component.scss
```diff
@@ -0,0 +1,18 @@
+textarea {
+  width: 80% !important;
```
JonLazariniDD, line 2: why are we overriding all the text area width to 80% ?

JonLazariniDD: Also it seems that we don't have this component so we can create some scss for it that we load in global.scss
We will have to remove all the hardcoded values, we can use the ones provided in the styleguide for the scaffolding elements such as color, font-size, font-weight.
If we don't have them we can create variables for them.

src/app/technical-record/test-history/test-history.component.ts
```diff
@@ -4,6 +4,7 @@ import { TestResultModel } from '@app/models/test-result.model';
 @Component({
   selector: 'vtm-test-history',
   templateUrl: './test-history.component.html',
+  styleUrls: ['../../app.component.scss'],
```
JonLazariniDD, line 7: why we loading the app.scss in here?
We shouldn't be having scoped css as agreed.

src/app/test-record/select-test-type/select-test-type.component.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+  }));
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(SelectTestTypeComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
```
JonLazariniDD, line 43: Add a snapshot please

src/app/test-record/test-record.mapper.ts
```diff
@@ -29,6 +34,23 @@ export class TestRecordMapper {
     }
   }
 
+  getReasonsForAbandoning(vehicleType: string, testTypeId: string) {
```
JonLazariniDD, line 37: Can we test this function please?
If we plan to reuse tirTestTypeIds anywhere else we could move them into a constant file or where we declared the test-record enums

src/app/test-record/test-section/edit/test-section-edit.component.scss
```diff
@@ -0,0 +1,13 @@
+.govuk-link,
```
JonLazariniDD, line 1: we can remove that, we are overriding the entire govuk class here

src/app/test-record/test-section/edit/test-section-edit.component.scss
```diff
@@ -0,0 +1,13 @@
...
+.govuk-link:visited {
+  color: #1d70b8 !important;
+  font-weight: 500;
+  cursor: pointer;
+  text-decoration: underline;
+  line-height: 2rem;
+  float: right;
+}
+
+.govuk-checkboxes__label {
```
JonLazariniDD, line 11: If we have to tweak the text because the content is not wrapping, let's create our own util class and not override the existing lib

src/app/test-record/test-section/edit/test-section-edit.component.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+    component = fixture.componentInstance;
+    component.testRecord = TESTING_TEST_MODELS_UTILS.mockTestRecord();
+    component.testType = TESTING_TEST_MODELS_UTILS.mockTestType();
+    component.applicableTestTypeIds1 = { 43: 'test' };
+    component.applicableTestTypeIds2 = { 43: 'test' };
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
```
JonLazariniDD, line 40: Need a snapshot please

src/app/test-record/test-section/edit/test-section-edit.component.ts
```diff
@@ -0,0 +1,158 @@
...
+    private dialog: MatDialog,
+    private router: Router,
+    parentForm: FormGroupDirective,
+    private testRecordMapper: TestRecordMapper
+  ) {
+    this.testResultChildForm = parentForm;
+  }
+
+  ngOnInit() {
+    this.isAbandoned = this.testType.testResult === 'abandoned';
```
JonLazariniDD, line 56: I think this could be moved into a service so that we remove any logic in there and make it testable

src/app/test-record/test-section/edit/test-section-edit.component.ts
```diff
@@ -0,0 +1,158 @@
...
+    this.reasonsForAbandoning = this.testRecordMapper.getReasonsForAbandoning(
+      this.testRecord.vehicleType,
+      this.testType.testTypeId
+    );
+    this.testTypeGroup = this.testResultChildForm.form.get('testType') as FormGroup;
+
+    if (!!this.testTypeGroup) {
```
JonLazariniDD, line 63: This can also be moved into the service and also refactored to render elements dynamically with computed values.
In fact this could be refactored like that i think.
```
const createElements = (listOfFields) => 
  listOfFields
    .map(field => 
       this.testTypeGroup.addControl(
          field, 
         new FormControl(this.testType[field], Validators.required)
        )
)

createElements(['testResult', 'reasonForAbandoning', ....])
```

src/app/test-record/select-test-type/select-test-type.component.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+      declarations: [SelectTestTypeComponent],
+      providers: [
+        {
+          provide: Store,
+          useValue: {
+            dispatch: jest.fn(),
+            pipe: jest.fn(() => hot('-a', { a: INITIAL_STATE })),
+            select: jest.fn()
+          }
+        },
```
alleypa, line 24: a component file should not be having references to the store

src/app/test-record/select-test-type/select-test-type.component.ts
```diff
@@ -0,0 +1,24 @@
...
+
+@Component({
+  selector: 'vtm-select-test-type',
+  templateUrl: './select-test-type.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class SelectTestTypeComponent implements OnInit {
+  testRecordObservable$: Observable<TestRecordTestType>;
+
+  constructor(private store: Store<IAppState>, private route: ActivatedRoute) {}
```
alleypa, line 17: hmmmm?

src/app/test-record/test-record.mapper.ts
```diff
@@ -167,6 +190,23 @@ export class TestRecordMapper {
     return !!countryCode ? countryCode : '';
   }
 
+  returnDateTimeISOString(dateString: string) {
```
alleypa, line 193: please lets talk about this

src/app/test-record/test-record.mapper.ts
```diff
@@ -220,6 +261,31 @@ export class TestRecordMapper {
...
+      testResultFormData.testType.reasonForAbandoning,
+      this.getReasonsForAbandoning(testResultMapped.vehicleType, testTypeMapped.testTypeId)
+    );
+    testTypeMapped.additionalCommentsForAbandon =
+      testResultFormData.testType.additionalCommentsForAbandon;
+    testTypeMapped.certificateNumber = testResultFormData.testType.certificateNumber;
+    testTypeMapped.testExpiryDate = testResultFormData.testType.testExpiryDate;
+    testTypeMapped.testAnniversaryDate = testResultFormData.testType.testAnniversaryDate;
+
+    testTypeMapped.testTypeStartTimestamp = this.returnDateTimeISOString(
```
alleypa, line 275: we should not manually construct date strings

src/app/test-record/test-section/edit/test-section-edit.component.ts
```diff
@@ -0,0 +1,188 @@
...
+    }
+  }
+
+  mapReasonsToFormGroup(options: SelectOption[]) {
+    return options.map((option) => {
+      return new FormControl(option.selected);
+    });
+  }
+
+  splitDateTimeString(dateTimeString: string, separator: string): string[] {
```
alleypa, line 166: your built control should be doing this

src/app/test-record/test-section/edit/test-section-edit.component.ts
JonLazariniDD: In fact I have noticed that we may want to create different things so we can still use that function and apply composition to it.
```
const withSpecialElementForm = (ctrl) => (listOfFields) => .... // this will return createElements() accepting now ctrl arg which will instantiate what we want
```

PR #209, opened 2020-06-17
feat(10253): autocomplete on make field

rspasici:
added autocomplete field on Make of Hgv creation section  https://jira.dvsacloud.uk/browse/CVSB-10253


src/app/shared/components/autocomplete/autocomplete.component.html
```diff
@@ -1,11 +1,12 @@
 <input
   class="govuk-input autocomplete-input"
+  ngClass="govuk-input govuk-input--width-{{ govInputLength }}"
```
alleypa, line 3: `govuk-input` you dont need this extra one

src/app/technical-record/body/edit/body-edit.component.html
```diff
@@ -8,13 +8,14 @@
...
               Make
             </label>
-            <input
-              id="test-edit-make"
-              type="text"
-              maxlength="30"
+            <vtm-autocomplete
+              [autocompleteData]="makeOptions"
+              [govInputLength]="20"
+              id="make"
```
alleypa, line 14: use this `test-edit-make`

src/global.scss
```diff
@@ -1,4 +1,4 @@
-@import 'node_modules/govuk-frontend/govuk/all';
+@import "node_modules/govuk-frontend/govuk/all";
```
alleypa, line 1: let catch up on this file @rspasici. Seems your editor is inserting double quotes

src/app/shared/components/autocomplete/autocomplete.component.scss
```diff
@@ -0,0 +1,3 @@
+.autocomplete-input::-webkit-calendar-picker-indicator {
```
JonLazariniDD, line 1: let's move that into _global.scss and remove this file and import please 
```(styleUrls: ['./autocomplete.component.scss'],)```
It is best to rely on plain sass/css stylesheet than creating a dependency with a framework and inlined styling that is difficult to port over.

File: src/global.scss [hunk 5]
```diff
@@ -31,7 +31,7 @@
 // Typography
 
 .font--green {
-  color: govuk-colour('green');
+  color: govuk-colour("green");
```
JonLazariniDD, line 34: it would be good if we could use the sass variables given by for "green" or "white" so we don't hardcode the values.

https://design-system.service.gov.uk/styles/colour/
(Colour palette section)

src/app/shared/components/autocomplete/autocomplete.component.scss
alleypa: @JonLazariniDD  i suggested this approach for @rspasici , as an update to do. It will make the autocomplete complete self contained and with its own single responsibility. The extra css added should not be global. I think it should be left in place in `styleUrls: ['./autocomplete.component.scss']`

JonLazariniDD: When I said in the global scss i meant loaded as a separate entity (It could be a partial) for example or in there as the file is quite small and managed to use the css lib really well plus this autocomplete component is a reusable component. 
I am against having inlined/scoped css especially when we use a css lib (which is plain css). If for some some reason it was decided to switch from Angular to React (or any other js framework in the future), this style would be lost as it is part/coupled to a component now whereas if it was contained within a compiled stylesheet, we could simply load the css file (containing all the styling) in the new framework and use the same html markup but wrapped within whatever new Js framework existing.
I think the css is pretty tidy now, the govuk lib pretty clear and I don't see the real benefit of scoping css into components.

JonLazariniDD:
We can add a description and link to the ticket as well please

JonLazariniDD:
Looks good.
We can check that the IDs are fine with testers and run the e2e test suite if not done yet please. Testers can help if we need assistance.

src/app/shared/components/autocomplete/__snapshots__/autocomplete.component.spec.ts.snap
```diff
@@ -40,11 +40,13 @@ exports[`AutocompleteComponent should populate the internal input when an extern
       ngdefaultcontrol=""
     >
       <input
-        class="govuk-input autocomplete-input ng-untouched ng-pristine ng-valid"
+        class="govuk-input autocomplete-input govuk-input--width- ng-untouched ng-pristine ng-valid"
```
JonLazariniDD, line 43: this not a valid css class I think `govuk-input--width-` is a number missing?

src/app/shared/components/autocomplete/autocomplete.component.html
```diff
@@ -1,11 +1,12 @@
...
   type="text"
-  name="{{id}}"
+  name="{{ id }}"
   id="test-{{ id }}"
-  [attr.list]="'autocomplete-'+id"
+  [attr.list]="'autocomplete-' + id"
   [formControl]="autocompleteInput"
 />
-<datalist id="{{'autocomplete-'+id}}">
+<datalist id="{{ 'autocomplete-' + id }}">
```
JonLazariniDD, line 10: needs to be prepended with `test-` please

src/app/shared/components/autocomplete/autocomplete.component.scss
JonLazariniDD: Why did we resolve that?

JonLazariniDD:
We will need to discuss about the scroll behaviour.

PR #211, opened 2020-06-19
Feature/cvsb 15867 - Save and Cancel HGVs

alleypa:
https://jira.dvsacloud.uk/browse/CVSB-15867

    CVSB-15867
    - Add implementation to save created vehicle record
    - Add directive implementation to make control nullable
    - Add implementation and tests for create effects


JonLazariniDD:
Looks good.
Let's add the ticket number for reference, check that the IDs are fine with testers and run the e2e test suite if not done yet please. We can check with testers if required if we need assistance.

src/app/store/actions/VehicleTechRecordModel.actions.spec.ts
```diff
@@ -1,81 +1,135 @@
...
+  techRecord: [mockTechRecord]
+} as VehicleTechRecordModel;
+
+const mockVehicleRecordEdit: VehicleTechRecordEdit = {
+  primaryVrm: 'vrm1',
+  techRecord: [mockTechRecord]
+} as VehicleTechRecordEdit;
+
+describe('GetVehicleTechRecordHavingStatusAll', () => {
+  it('should create GetVehicleTechRecordHavingStatusAll action', () => {
```
JonLazariniDD, line 35: We could be adding with the correct payload as this is what this action creator does so it exactly describes what this is about.
it takes a payload in that will then be consumed in the reducer.

src/app/store/actions/VehicleTechRecordModel.actions.spec.ts
```diff
@@ -1,81 +1,135 @@
...
-      type: EVehicleTechRecordModelActions.GetVehicleTechRecordModel,
-      payload: techRecord
+      type: EVehicleTechRecordActions.GetVehicleTechRecordHavingStatusAll,
+      payload: searchParams
     });
+  });
+});
+
+describe('GetVehicleTechRecordHavingStatusAllSuccess', () => {
+  it('should create GetVehicleTechRecordHavingStatusAllSuccess action', () => {
```
JonLazariniDD, line 47: could be doing the same here so it seems to be taking something as a param and generate an action with a type and the given/provided payload.

src/app/store/effects/VehicleTechRecordModel.effects.spec.ts
```diff
@@ -1,177 +1,260 @@
...
 import { SearchParams } from '@app/models/search-params';
 import { SetErrorMessage } from '@app/store/actions/Error.actions';
 
-const techRecordModel: VehicleTechRecordModel = {
-  systemNumber: '1231243',
-  vrms: null,
-  vin: 'ABCDEFGH777777',
-  techRecord: [],
-  metadata: { adrDetails: undefined }
+const mockVehicleTechRecord = (): VehicleTechRecordModel => {
```
JonLazariniDD, line 33: Is that something we may want to be moving into our test util file?
We already know that it should be used as well by Test result.

src/app/store/actions/VehicleTechRecordModel.actions.spec.ts
alleypa: @JonLazariniDD  what action does mainly is to create `an instance` and the output should be inferred in the assertion i.e. creating the correct type with or without payload. Specifying the creation of the `correct payload in the title` does not necessarily convey much meaning as the encapsulated info is of less importance at this stage but just the type which again the tests documents and asserts

alleypa:
rebase against 00000

src/app/store/actions/VehicleTechRecordModel.actions.spec.ts
JonLazariniDD: What I was trying to point out is that it is not explicitly stating what the redux (elm) architecture is about.
This particular action creator takes a payload (searchParams) as an argument and returns a new object containing this payload and its relevant action type bound to it. We are not capturing the 'moving part' of this action creator which is the payload I think. Also, probably down to personal preference but I am not a fan of having the payload property in the action as payload as it is not explicit I find, searchParams has more meaning than payload if we had to follow this action through the codebase. It is the convention we have for now and can stick with it.
So up to you how you wish to update or not but here we could be assuming that developers would have a knowledge of the flux/redux architecture and if I only was to read the description of the this particular test I wouldn't know what it does without looking at the implementation.
This could be overthought from me but that was my intent behind this comment.

PR #203, opened 2020-06-11
Feature/cvsb 10184 modal popup general implementation

m-ujaffer:
This functional and does the job but is not covering the tech debt to replace the functionality from test-records.

PR #229, opened 2020-06-29
feat(cvsb-10184): lose changes modal  with common logic for all modals

m-ujaffer:
## Lose Changes Modal
Lose changes modal for the overall application and common code for all modals in general.
original PR
https://github.com/dvsa/cvs-app-vtm/pull/203
https://jira.dvsacloud.uk/browse/CVSB-10184

## Checklist

- [X] Branch is rebased against the latest develop/common
- [X] Necessary `id` required prepended with `"test-"` have been checked with automation testers and added
- [X] Code and UI has been tested manually after the additional changes
- [X] PR title includes the JIRA ticket number
- [X] Squashed commits contain the JIRA ticket number
- [X] Link to the PR added to the repo


src/app/directives/form-connector/form-connector.directive.ts
```diff
@@ -16,16 +20,22 @@ export class FormConnectorDirective implements OnInit, OnDestroy {
...
+        tap(([formValue, appForm]: [FormGroup, boolean]) => {
+          console.log('VALUE: ', formValue, 'PATH: ', this.path);
+          if (!this.formGroupDirective.form.pristine && appForm) {
+            this.store.dispatch(new SetAppFormDirty());
+          }
+        })
       )
       .subscribe();
-      console.log(this.formGroupDirective.form);
+    console.log(this.formGroupDirective.form);
```
JonLazariniDD, line 38: Why are we keeping this console log?
Is that for debug purpose?
If it is the case can we please add a TODO Add flag so we enable that with a flag.
(CVSB-15050)
I have a branch with the changes for that but I didn't get a change yet to deploy and run the tests yet however we could enable that with a flag once this PR gets in.

src/app/guards/form-state.guard.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+    }).compileComponents();
+    formStateGuard = TestBed.get(FormStateGuard);
+    mockSelector.next({
+      getAppFormState: false
+    });
+  }));
+
+  describe('canActivate', () => {
+    it('should call LoadModal action', () => {
+      // TODO: update unit test
```
JonLazariniDD, line 35: Is that still needed?

src/app/modal/components/lose-changes/lose-changes.component.scss
```diff
@@ -0,0 +1,29 @@
+@import "node_modules/govuk-frontend/govuk/all";
+
+.mat-dialog-content {
+  overflow: hidden;
+}
+
+.title {
+  color: #0B0C0C;
```
JonLazariniDD, line 8: Can we use a color from govuk instead please?
$govuk-text-colour should work

(for reference; https://design-system.service.gov.uk/styles/colour/)

src/app/modal/components/lose-changes/lose-changes.component.scss
```diff
@@ -0,0 +1,29 @@
...
+  font-size: 36px;
```
JonLazariniDD, line 9: can this work if we use these classes? https://design-system.service.gov.uk/styles/typography/

src/app/modal/components/lose-changes/lose-changes.component.scss
```diff
@@ -0,0 +1,29 @@
...
+  font-weight: 700;
+  line-height: 43px;
+  width: 705px;
```
JonLazariniDD, line 12: Does it need to be responsive?
If yes we could use some relative size and use max-width, min-width to lock the ratio

src/app/modal/components/lose-changes/lose-changes.component.scss
```diff
@@ -0,0 +1,29 @@
...
+  text-align: left;
+}
+
+.cancel {
+  color: #005EA5;
```
JonLazariniDD, line 17: Couldn't find this colour defined, i assume this is the greyish secondary colour?

Maybe we can declare it if it is not present
I assume the CTA will look like this?

image



govuk-colour("light-grey") could be what we wanted perhaps

src/app/modal/modal.component.scss
```diff
@@ -1,6 +1,6 @@
 @import "node_modules/govuk-frontend/govuk/all";
 
-.landing-page {
+.modal-page {
 
   & a {
     color: #1d70b8 !important;
```
JonLazariniDD, line 6: 😖 something to deal with later but i will check how this got in, !important is nasty

src/app/directives/form-connector/form-connector.directive.ts
m-ujaffer: Yes, it is for debugging purpose and would be removed eventually as it is used by Olu and others to debug formGroup. TODO comment is added.

src/app/modal/components/lose-changes/lose-changes.component.scss
m-ujaffer: removed the title class. It slightly differs the design but this is something that we can agree with testers and BAs.

src/app/modal/components/lose-changes/lose-changes.component.scss
m-ujaffer: This is actually looks like a link. I tried using the `govuk-link` class. It was behaving properly. I've updated the css to use `$govuk-link-colour`.

src/app/modal/modal.component.scss
m-ujaffer: removed the css file

src/app/directives/form-connector/form-connector.directive.ts
JonLazariniDD: nice 👍, got a PR for this https://github.com/dvsa/cvs-app-vtm/pull/232
Trying to see with the devOps how to get this through as the config has been changed .This will require pipeline changes (very likely)

src/app/modal/components/lose-changes/lose-changes.component.html
```diff
@@ -0,0 +1,14 @@
...
+    </span>
+  </div>
+
+  <button id="test-lose-changes" class="govuk-button" data-module="govuk-button" (click)="onLoseChangesClick()" cdkFocusInitial>
+    Leave and lose changes
+  </button>
+
+  <a  id= "test-cancel" class="cancel" (click)="onCancelClick()">
+    <span>Continue changing record</span>
+  </a>
```
alleypa, line 14: this will work with some tweaks @m-ujaffer. we can play with the govs margin or padding to look accurate. Hence we wont need the custom css file

`<span
            id="test-cancel"
            class="govuk-!-font-size-19 govuk-!-margin-left-5"
          >
            <a
              href="javascript:void(0)"
              class="govuk-link govuk-link--no-visited-state"
              (click)="onCancelClick()"
            >
              <span>Continue changing record</span>
            </a>
          </span>`

PR #242, opened 2020-07-06
fix(cvsb-10184): handle header for low page width --- [dep fix(cvsb-17179)]

m-ujaffer:
## Header bug fix and modification in AppFormDirty.
Header bug is fixed to render properly on smaller screen size. FormConnector is not used for triggering AppFormDirty. Instead ViewState is used.
Bug fix for back button behavior when test record is in edit mode.
[https://jira.dvsacloud.uk/browse/CVSB-10184]()
https://jira.dvsacloud.uk/browse/CVSB-17156

## Checklist

### DIFF
--- a/src/app/store/effects/app-form.effects.spec.ts
+++ b/src/app/store/effects/app-form.effects.spec.ts
@@ -8,8 +8,9 @@ import { SetViewState } from '../actions/VehicleTechRecordModel.actions';
 import { VIEW_STATE } from '@app/app.enums';
 
 import { AppFormEffects } from './app-form.effects';
-import { SetAppFormPristine } from '../actions/app-form-state.actions';
+import { SetAppFormPristine, SetAppFormDirty } from '../actions/app-form-state.actions';
 import { ClearErrorMessage } from '../actions/Error.actions';
+import { SetTestViewState } from '../actions/VehicleTestResultModel.actions';
 
 describe('AppFormEffects', () => {
   let effects: AppFormEffects;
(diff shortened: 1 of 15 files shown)


src/app/store/effects/app-form.effects.spec.ts
```diff
@@ -29,7 +30,7 @@ describe('AppFormEffects', () => {
     expect(effects).toBeTruthy();
   });
 
-  describe('setAppFormPristine$', () => {
+  describe('setTechFormPristine$', () => {
     it('should dispatch SetAppFormPristine action if view state is set to view only', () => {
```
alleypa, line 34: `should dispatch SetAppFormPristine` use `setTechFormPristine`

alleypa, line 34: use `should dispatch SetAppFormPristine & ClearErrorMessage actions if view state is set to view only`

src/app/store/effects/app-form.effects.spec.ts
```diff
@@ -48,7 +49,74 @@ describe('AppFormEffects', () => {
...
+      actions$ = hot('-a--', { a: action });
+
+      const expected$ = cold('--');
+
+      expect(effects.setTechFormDirty$).toBeObservable(expected$);
+    });
+  });
+
+  describe('setTestFormPristine$', () => {
+    it('should dispatch SetAppFormPristine action if test view state is set to view only', () => {
```
alleypa, line 79: similar here use `should dispatch SetAppFormPristine & ClearErrorMessage actions if test view state is set to view only`

PR #286, opened 2022-04-21
CB2-3314: Global Error Handling

tom-bjss:
A new global error component for displaying errors on the services, this is an MVP and a second iteration will be done to map errors to fields to comply with GDS.


src/app/store/test-records/reducers/test-records.reducer.spec.ts
```diff
@@ -1,4 +1,4 @@
-import { mockTestResult, mockTestResultList } from '../../../../mocks/mock-test-result';
```
NiallBunting, line 1: Is this being used?

NiallBunting (approved):

src/app/store/test-records/reducers/test-records.reducer.spec.ts
tom-bjss: No, all the tests use the mock list

andrewf-bjss (approved):

PR #287, opened 2022-04-22
feat(cb2-3218): view 5 additional fields for annual psv

andrewf-bjss:
## View 5 Additional fields for an Annual PSV Test Result

- Adds 5 additional data fields from the annual PSV test result
- Sets the foundation for dynamic tech (by vehicle) and test (by test-type) fields, using Dynamic Forms

[https://jira.dvsacloud.uk/browse/CB2-3218](CB2-3218)

## Checklist

- [x] Delete branch after merge


naathanbrown (approved):

src/app/store/technical-records/technical-record-service.reducer.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+      };
+      const action = getByVINSuccess({ vehicleTechRecords: [...records] });
+      const state = vehicleTechRecordReducer(initialState, action);
+
+      expect(state).toEqual(newState);
+      expect(state).not.toBe(newState);
+    });
+  });
+
+  describe('getByVINSuccess', () => {
```
aaron-downing, line 32: can these two tests go within the same describe block? or has the describe block been copied and the description needs updating?

src/app/forms/components/dynamic-form-group/dynamic-form-group.component.ts
```diff
@@ -35,7 +35,9 @@ export class DynamicFormGroupComponent implements OnInit {
     // NiallB: This populates data without respecting where in the hierarchy it is (see unit tests)
     Object.keys(data).forEach((dataKey) => {
       if (typeof data[dataKey] === 'object') {
-        this.populateData(data[dataKey], template);
+        if (data[dataKey] !== null) {
```
tom-bjss, line 38: Can we use a truthy evaluation here (can only think of empty string being the issue)?

Also would a similar thing happen in the else block, is this explicitly data[datakey] check or more so on the data key?

Also if the above doesn't apply could this be added on the if statement above as to not reduce our nesting ?

aaron-downing: This is basically checking that if we had the following:
```
{
  someProperty: 'value'
  somePropertyObject: {}
}
```
...that `somePropertyObject` isn't actually empty because if it is empty we don't want to iterate inside it. So it might be worth adding it to the if statement above like you mentioned. 

The else statement is concerned for when the value of `data[datakey]` isn't an object i.e `someProperty` so I don't believe we would need a null check like we do in the first if statement as we're not looking to iterate over it. If the value of `someProperty` is null then we would only be putting the value of null within `template.value` (which is already `null`/`""`) so it wouldn't really be worth adding a null check here.

aaron-downing (approved):

src/app/forms/components/dynamic-form-group/dynamic-form-group.component.ts
andrewf-bjss: @tom-bjss & @aaron-downing 
We found we see
```
{
  someProperty: null
}
```

Where `null` passes the `=== object` test but `null` itself doesn't have any keys (so `Object.keys(null).forEach(...)` obviously fails)

But yes, we could do `if (data[dataKey]) { /* do stuff */ }` as a truthy-style guard clause, I just thought the ` !== null` was explicitly clear. Happy to change it  :)

tom-bjss (approved):

src/app/features/home/home.component.ts
```diff
@@ -1,10 +1,17 @@
...
+import { FormBuilder, FormGroup, Validators } from '@angular/forms';
 
 @Component({
   selector: 'app-home',
   templateUrl: './home.component.html',
   styleUrls: ['./home.component.scss']
 })
 export class HomeComponent {
-  constructor() {}
+  form: FormGroup;
```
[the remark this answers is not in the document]
andrewf-bjss: Nope, removed  👍🏻

PR #295, opened 2022-04-29
feature/CB2-3618: Loading spinner

NiallBunting:
## Ticket title

_One line description_
[link to ticket number]()

## Checklist

### DIFF
--- a/mock-api/mock-api-server.ts
+++ b/mock-api/mock-api-server.ts
@@ -10,7 +10,7 @@ const args = require('minimist')(process.argv.slice(2));
 server.use(middlewares);
 
 // Add custom routes before JSON Server router
-server.get('/vehicles/*', (req, res) => {
+server.get('/vehicles/:id/*', (req, res) => {
   switch (args['tech-record']) {
     case 'NotFound':
       res.status(404);
@@ -23,7 +23,15 @@ server.get('/vehicles/*', (req, res) => {
       res.jsonp('Error service unavailable');
       break;
     default:
-      res.jsonp(mockVehicleTecnicalRecordList());
+      // Switch on VIN
+      switch (req.params.id) {
+        case 'delay':
+          console.log('Delaying request');
+          setTimeout(() => {res.jsonp(mockVehicleTecnicalRecordList())}, 2500);
+          break;
+        default:
+          res.jsonp(mockVehicleTecnicalRecordList());
+      }
       break;
   };
 });
--- a/mock-api/mock-api-server.ts
+++ b/mock-api/mock-api-server.ts
@@ -5,54 +5,48 @@ const server = jsonServer.create();
 const router = jsonServer.router('{}');
 const middlewares = jsonServer.defaults();
 
-const args = require('minimist')(process.argv.slice(2));
-
 server.use(middlewares);
 
 // Add custom routes before JSON Server router
 server.get('/vehicles/:id/*', (req, res) => {
-  switch (args['tech-record']) {
-    case 'NotFound':
-      res.status(404);
-      res.statusMessage = 'NotFound';
-      res.jsonp('Error no vehicle found');
+  // Switch on VIN
+  switch (req.params.id) {
+    case 'delayok':
+      console.log('Delaying request');
+      setTimeout(() => {res.jsonp(mockVehicleTecnicalRecordList())}, 2500);
       break;
-    case 'ServiceError':
-      res.status(500);
-      res.statusMessage = 'Unavailable';
-      res.jsonp('Error service unavailable');
+    case 'delayservererror':
+      console.log('Delaying not found request');
+      setTimeout(() => {
+        res.status(500);
+        res.statusMessage = 'Unavailable';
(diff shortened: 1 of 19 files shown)


src/app/layout/spinner/spinner.component.html
```diff
@@ -0,0 +1,22 @@
+<div class="overlay" *ngIf="(spinnerService.showSpinner$ | async) == true">
+  <div class="govuk-box spinner-wrapper" aria-busy="true">
+    <div id="loader" class="govuk-c-loader" role="progressbar" aria-label="Loading, please wait." tabindex="0">
+      <div class="spinner-container">
+        <div class="spinner">
+          <div></div>
```
tom-bjss, line 6: What are all these divs for ?

NiallBunting: Came from the old implementation, so unsure but just copied it as is.

src/app/layout/spinner/spinner.service.ts
```diff
@@ -0,0 +1,24 @@
...
+import { select, Store } from '@ngrx/store';
+import { State } from '@store/.';
+import { spinnerState, SpinnerState } from '@store/spinner/reducers/spinner.reducer';
+import { Subject, combineLatest, map, Observable } from 'rxjs';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class SpinnerService implements OnDestroy {
+  private readonly _destroying$ = new Subject<void>();
```
danielchildsbjss, line 11: Is this needed?

tom-bjss (dismissed):

src/app/features/global-error/global-error.component.spec.ts
```diff
@@ -8,7 +8,7 @@ import { GlobalErrorService } from './global-error.service';
 
 @Component({
   selector: 'app-mock-component',
-  template: '<app-global-error></app-global-error><input id="test-input" type="text" />',
+  template: '<app-global-error></app-global-error><input id="test-input" type="text" />\n',
```
mcadenas-bjss, line 11: Thanks! Couldn't figure out how to get rid of that lint warning.

src/app/layout/spinner/spinner.service.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+    mockStore = TestBed.inject(Store);
+
+    service = new SpinnerService(mockStore);
+  });
+
+  it('should create the user service', () => {
+    expect(service).toBeTruthy();
+  });
+
+  it('should get and set the username', (done) => {
```
mcadenas-bjss, line 31: Test name doesn't match test.

mcadenas-bjss (approved):

danielchildsbjss (dismissed):

mcadenas-bjss (approved):

mock-api/mock-api-server.ts
```diff
@@ -23,7 +23,23 @@ server.get('/vehicles/*', (req, res) => {
...
+          console.log('Delaying not found request');
+          setTimeout(() => {
+            res.status(500);
+            res.statusMessage = 'Unavailable';
+            res.jsonp('Error service unavailable');
+          }, 2500);
+          break;
+        default:
+          res.jsonp(mockVehicleTecnicalRecordList());
+      }
```
andrewf-bjss, line 42: I feel like we'd started down a path of command-line switches to affect the mock-api responses, whereas this approach introduces magic search criteria. Could we stick to one approach for simplicity's sake?

Personally I like the command-line approach, but I don't mind the change if there's a compelling reason.

Staying command-line, we could wrap the entire switch statement in a `setTimeout(() => {...}, getDelay())` where `getDelay()` returns the value from the command line switch or return zero if not defined so the response is not delayed.

We could use this same approach for all of the mocked api calls (i.e. test-records), and it keeps the NotFound, and ServiceError logic to a single implementation (the code as-is has 2 different HTTP responses for NotFound - a 404 (line 16) and 500 (line 35)).

@danielchildsbjss - what do you think?

andrewf-bjss (dismissed):

mock-api/mock-api-server.ts
NiallBunting: The reason I didn't do that is it's a pain to keep changing what you want from the api. It's much easier to leave the service running and just choose what you want. Eg. with the error message stuff with @gjulien-bjss it was much easier to just do `return404` and `return202` commands to check that the error messages were working as expected.

NiallBunting: Converted the entire PR to this method, awaiting discussion.

danielchildsbjss: I'm happy with this approach, but we might start hitting problems if/when we add validation to the VIN search box and these secret commands can't get passed through.
Some of these things I'm not sure we need - if we want to simulate 500 not available why not just close the mock server and make it really unavailable?

mcadenas-bjss (approved):

tom-bjss (approved):

PR #296, opened 2022-04-29
feat(cb2-2890): create and display tech vehicle summary

naathanbrown:
## Ticket title

Create a vehicle summary from the tech record attached to the vehicle.
[link to ticket number](https://jira.dvsacloud.uk/browse/CB2-2890)

## Checklist

- [ x ] Branch is rebased against the latest develop/common
- [ x ] Code and UI has been tested manually after the additional changes
- [ x ] PR title includes the JIRA ticket number
- [ x ] Squashed commits contain the JIRA ticket number
- [ x ] Link to the PR added to the repo
- [ x ] Delete branch after merge


src/app/forms/templates/hgv/hgv-tech-record.template.ts
```diff
@@ -0,0 +1,153 @@
...
+            type: FormNodeTypes.CONTROL,
+            viewType: FormNodeViewTypes.VEHICLETYPE
+        },
+        {
+            name: 'regnDate',
+            label: 'Date of first registration',
+            value: '',
+            children: [],
+            type: FormNodeTypes.CONTROL,
+            viewType: FormNodeViewTypes.STRING
```
tom-bjss, line 22: With the default type above all of these string types can be removed

src/app/models/vehicle-tech-record.model.ts
```diff
@@ -1,11 +1,134 @@
...
+  l6e = 'l6e',
+  l7e = 'l7e'
+}
+
+export enum VehicleSizes {
+  SMALL = 'small',
+  LARGE = 'large'
+}
+
+export interface TechRecordModel {
```
danielchildsbjss, line 98: Do we want a smarter model than this? Some of these fields are not applicable for certain vehicle types, do we want to model each vehicle type as its own class? They can all share a common base class

src/app/features/tech-record-summary/tech-record-summary.component.ts
```diff
@@ -0,0 +1,62 @@
...
+  ngOnInit(): void {
+    this.vehicleTemplate();
+    this.currentRecord = this.viewableTechRecord(this.vehicleTechRecord)
+  }
+
+  constructor() {
+  }
+
+  viewableTechRecord(record: VehicleTechRecordModel): TechRecordModel | undefined {
+    let viewableTechRecord = record?.techRecord?.find(record => record.statusCode === StatusCodes.PROVISIONAL)
```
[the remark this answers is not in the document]
naathanbrown: A conversation with Steven Pears, however it is worth following up with the PO. Have added a comment to the ticket to have this checked but he is confident that would be the way

src/app/models/vehicle-tech-record.model.ts
aaron-downing: I think we're kind of using this as a generic 'what we can expect back from the DynamoDB'. But I guess in the get request to the DB you could specify the return type for the response to be something like `<PsvTechRecordModel | HgvTechRecordModel | TrlTechRecordModel | ...>` but maybe that's a bit too much? Especially when we add additional vehicle types.

Our vehicle specific type 'classes' seem to be based in the templates we have for the dynamic forms where we have separate ones for PSV, HGV and TRL.

src/app/vehicle-technical-record/vehicle-technical-record.component.ts
```diff
@@ -10,7 +10,7 @@ import { VehicleTechRecordModel } from '../models/vehicle-tech-record.model';
   styleUrls: ['./vehicle-technical-record.component.scss']
 })
 export class VehicleTechnicalRecordComponent {
-  @Input() vehicleTechRecord?: VehicleTechRecordModel;
+  @Input() vehicleTechRecord!: VehicleTechRecordModel;
```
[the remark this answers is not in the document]
aaron-downing: I assumed that we would only be rendering this component if we had retrieved a tech record back from the DB during the search in which case the component wouldn't expect the input to be undefined?

src/app/forms/templates/hgv/hgv-tech-record.template.ts
naathanbrown: Sorry, what do you mean? I am confused

tom-bjss: My above comment about the default switch on the view item list component. It means if you don't supply the view type it will default to String

src/app/features/tech-record-summary/tech-record-summary.component.ts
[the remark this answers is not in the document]
danielchildsbjss: We do need to be careful about implementing logic that's not in the user story as it's going to make things difficult to test and verify

src/app/forms/templates/hgv/hgv-tech-record.template.ts
danielchildsbjss: I'm okay if we leave them in and be explicit

src/app/features/tech-record-summary/tech-record-summary.component.ts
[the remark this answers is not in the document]
naathanbrown: I'm happy to remove it but the current ticket doesn't then specify which should be used, so we'll need to make an assumption there too (assume current but). I'll do whatever you think is best :)

src/app/models/vehicle-tech-record.model.ts
danielchildsbjss: I think `<PsvTechRecordModel | HgvTechRecordModel | TrlTechRecordModel | ...>` will be exactly what we need, especially if we want type safety and logical consistency. What comes from DynamoDb doesn't have to define our classes in Angular (we can always map the data coming back to new types). But maybe this is something we can build later.

src/app/features/tech-record-summary/tech-record-summary.component.ts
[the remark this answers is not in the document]
danielchildsbjss: Let's leave it for now if the logic has been written, just things to keep in the back of our mind for the future. We want to get these things figured out in refinement and called out in the ticket.

src/app/features/tech-record-summary/tech-record-summary.component.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { TechRecordModel, VehicleTypes, StatusCodes, FuelTypes, VehicleConfigurations, EuVehicleCategories, VehicleSizes } from '@models/vehicle-tech-record.model';
+import { By } from '@angular/platform-browser';
+
+
+describe('TechRecordSummaryComponent', () => {
+    let component: TechRecordSummaryComponent;
+    let fixture: ComponentFixture<TechRecordSummaryComponent>;
+
+    const fakeTechRecord: TechRecordModel = {
```
mcadenas-bjss, line 11: Create functions using `createMock` like in `src/mocks/mock-test-result.ts`. Saves time and keeps test readable.

src/app/features/tech-record-summary/tech-record-summary.component.ts
```diff
@@ -0,0 +1,62 @@
...
+    if (viewableTechRecord == undefined){
+      viewableTechRecord = record?.techRecord?.find(record => record.statusCode === StatusCodes.CURRENT)
+    }
+    if (viewableTechRecord == undefined){
+      viewableTechRecord = record?.techRecord?.find(record => record.statusCode === StatusCodes.ARCHIVED)
+    }
```
mcadenas-bjss, line 34: Regardless of the conversation above. Why are both of these checking for undefined? if `viewableTechRecord` is undefined then the second condition will overwrite the first

naathanbrown: Yeah that's what we want. If it is not assigned when we find on provisional it will be undefined, so try current, if that is also unassigned try archived. Means if it finds it in current it won't be placed by archive.

mcadenas-bjss: Could have a single find and use the `OR` operator.

src/app/models/vehicle-tech-record.model.ts
```diff
@@ -1,11 +1,138 @@
 export interface VehicleTechRecordModel {
-  vrms: VrmModel[];
+  vrms: vrms[];
```
andrewf-bjss, line 2: nitpick: the array brackets denote plurality, so the 's' can be dropped from the model
```suggestion
  vrms: Vrm[];
```

andrewf-bjss: Looks like all of the additional interface declarations are plural, and I'd suggest they should be singular; consider:

`StatusCode.ARCHIVED`
vs
`StatusCodes.ARCHIVED`

tom-bjss (approved):

mcadenas-bjss (dismissed):

PR #308, opened 2022-05-09
fix(cb2-3776): Change to a string to match database return

naathanbrown:
## Ticket title

Move model and template to a string, instead of array.
[link to ticket number](https://dvsa.atlassian.net/browse/CB2-3776?atlOrigin=eyJpIjoiYjNiYWVhZGJlNTZmNDgwYTk3MDNkZjZiODNmYTlkNGUiLCJwIjoiaiJ9)

## Checklist

### DIFF
--- a/src/app/forms/templates/psv/psv-annual.template.ts
+++ b/src/app/forms/templates/psv/psv-annual.template.ts
@@ -176,17 +176,10 @@ export const PsvAnnual: FormNode = {
             },
             {
               name: 'reasonForAbandoning',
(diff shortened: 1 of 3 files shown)


danielchildsbjss (approved):
I think this should be an array at some point and not a . separated string, as it is in the database. It is always better to push hacks towards the boundaries - so we can be turning the array to a string and back when we read/write to the API. But we can go this way for now.

aaron-downing (approved):

PR #311, opened 2022-05-13
feat(cb2-3411): added seatbelts to test form

kpearmanbjss:
## View Seatbelt installation check section for a PSV test record

Added the seatbelt fields to the Test results Form. 
[CB2-3411](https://dvsa.atlassian.net/browse/CB2-3411)

## Checklist

### DIFF
--- a/src/app/forms/templates/psv/psv-annual.template.ts
+++ b/src/app/forms/templates/psv/psv-annual.template.ts
@@ -222,6 +222,34 @@ export const PsvAnnual: FormNode = {
                 { value: false, label: 'No' }
               ],
               children: []
+            },
+            {
+              name: 'seatbeltSection',
+              label: 'Seatbelts',
+              type: FormNodeTypes.SECTION,
+              children: []
+            },
+            {
+              name: 'seatbeltInstallationCheckDate',
+              label: 'Carried out during test',
+              value: '',
+              children: [],
+              type: FormNodeTypes.CONTROL
+            },
+            {
+              name: 'numberOfSeatbeltsFitted',
+              label: 'Number of seatbelts fitted',
+              value: '',
+              children: [],
+              type: FormNodeTypes.CONTROL
+            },
+            {
+              name: 'lastSeatbeltInstallationCheckDate',
+              label: 'Most recent installation check',
+              value: '',
+              children: [],
+              type: FormNodeTypes.CONTROL,
+              viewType: FormNodeViewTypes.DATE
             }
           ]
         }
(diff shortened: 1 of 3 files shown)


gjulien-bjss (approved):

src/app/models/test-type.model.ts
```diff
@@ -15,4 +15,8 @@ export interface TestType {
   prohibitionIssued: boolean;
 
   testResult: string;
+
+  seatbeltInstallationCheckDate: boolean;
```
Ian-Hodges, line 19: Should seatbeltInstallationCheckDate have 'Date' at the end? Sounds wrong when it's a boolean.

src/app/models/test-type.model.ts
```diff
@@ -15,4 +15,8 @@ export interface TestType {
...
+  numberOfSeatbeltsFitted: number;
+  lastSeatbeltInstallationCheckDate: string | Date;
```
danielchildsbjss, line 21: Is this the correct type? Feels like it should be a nullable Date

naathanbrown (approved):

src/app/models/test-type.model.ts
kpearmanbjss: I went off of what the other fields were using, I don't think we've had something that uses a nullable date yet so not entirely sure how I would change it to match that?

src/app/models/test-type.model.ts
kpearmanbjss: I was also confused, that is the way it's laid out on the DB according to the spreadsheet I was given, it's a boolean called seatbeltInstallationCheckDate, it shows whether or not the seatbelt checks were carried out during this test or not.

Ian-Hodges: Sounds like it is a boolean then, just badly named.

src/app/models/test-type.model.ts
kpearmanbjss: Looking at the data, the field appears to be "Were the seatbelts checked on this date" and if so it makes the other fields mandatory. I think lastSeatbeltInstallationCheckDate is a literal "The last time we checked the seatbelts".

I've gone on the assumption by the way these need to match the structure of how they are in the DB? If not then i'd happily rename it to something more useful but not sure how i'd point it to the right field

src/app/forms/templates/psv/psv-annual.template.ts
```diff
@@ -222,6 +222,34 @@ export const PsvAnnual: FormNode = {
...
+              name: 'numberOfSeatbeltsFitted',
+              label: 'Number of seatbelts fitted',
+              value: '',
+              children: [],
+              type: FormNodeTypes.CONTROL
+            },
+            {
+              name: 'lastSeatbeltInstallationCheckDate',
+              label: 'Most recent installation check',
+              value: '',
```
danielchildsbjss, line 249: I think then this also becomes null, if the control supports that

Ian-Hodges (approved):

PR #318, opened 2022-05-17
CB2-3871: Fixed Spinner On No Test Results

tom-bjss:
## https://dvsa.atlassian.net/browse/CB2-3871

The spinner would deploy when searching for a technical record but would not remain when searching for the test records.
This would then show the technical records but no test results which later would become populated without any notification to the user.

This change removes the spinner listening to the on success of get by vin, this is because the test results has an effect on the getByVinSuccess so its better to have that effect to reverse the state.

### DIFF
--- a/src/app/store/spinner/reducers/spinner.reducer.ts
+++ b/src/app/store/spinner/reducers/spinner.reducer.ts
@@ -1,6 +1,7 @@
 import { createFeatureSelector, createReducer, createSelector, on } from '@ngrx/store';
 import * as TechnicalRecordServiceActions from '../../technical-records/technical-record-service.actions';
 import * as TestResultActions from '@store/test-records';
+import { fetchTestResultsBySystemIdSuccess } from '@store/test-records';
 
 export const STORE_SPINNER_KEY = 'Spinner';
 
@@ -19,7 +20,7 @@ export const spinnerState = createSelector(getSpinnerState, (state) => state.sho
 export const spinnerReducer = createReducer(
   initialSpinnerState,
   on(TechnicalRecordServiceActions.getByVIN, TestResultActions.fetchTestResults, TestResultActions.fetchTestResultsBySystemId, (state) => ({ ...state, showSpinner: true })),
-  on(TechnicalRecordServiceActions.getByVINFailure, TechnicalRecordServiceActions.getByVINSuccess,
+  on(TechnicalRecordServiceActions.getByVINFailure,
      TestResultActions.fetchTestResultsBySystemIdFailed, TestResultActions.fetchTestResultsBySystemIdSuccess,
      TestResultActions.fetchTestResultsFailed, TestResultActions.fetchTestResultsSuccess,
       (state) => ({ ...state, showSpinner: false })


src/app/store/spinner/reducers/spinner.reducer.ts
```diff
@@ -1,6 +1,7 @@
 import { createFeatureSelector, createReducer, createSelector, on } from '@ngrx/store';
 import * as TechnicalRecordServiceActions from '../../technical-records/technical-record-service.actions';
 import * as TestResultActions from '@store/test-records';
+import { fetchTestResultsBySystemIdSuccess } from '@store/test-records';
```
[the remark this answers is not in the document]
kpearmanbjss: Also trying to work out how from the code shown it keeps the spinner for finishing early, unless this import has something to do with it ?

src/app/store/spinner/reducers/spinner.reducer.ts
```diff
@@ -19,7 +20,7 @@ export const spinnerState = createSelector(getSpinnerState, (state) => state.sho
 export const spinnerReducer = createReducer(
   initialSpinnerState,
   on(TechnicalRecordServiceActions.getByVIN, TestResultActions.fetchTestResults, TestResultActions.fetchTestResultsBySystemId, (state) => ({ ...state, showSpinner: true })),
-  on(TechnicalRecordServiceActions.getByVINFailure, TechnicalRecordServiceActions.getByVINSuccess,
+  on(TechnicalRecordServiceActions.getByVINFailure,
```
gjulien-bjss, line 23: Also the tests are broken for this reducer, since as things stand, the spec file says the test should fail but it doesn't. I've just sent over a message to Tom with a 'fix' for the spec file, hopefully it's done tomorrow

mcadenas-bjss: refactored

kpearmanbjss (approved):

src/app/layout/spinner/spinner.service.ts
```diff
@@ -1,18 +1,40 @@
...
 @Injectable({
   providedIn: 'root'
 })
 export class SpinnerService {
+  globalLoadingState$: Observable<boolean>;
+  testResultLoadingState$: Observable<boolean>;
+  techRecordsLoadingState$: Observable<boolean>;
 
   constructor(private store: Store) {
+    this.globalLoadingState$ = this.store.pipe(select(spinnerState));
```
danielchildsbjss, line 17: I like the moving of loading state back into the reducers that are triggering them, but what is changing the globalLoadingState?

src/app/layout/spinner/spinner.service.ts
```diff
@@ -19,7 +19,7 @@ export class SpinnerService {
     this.techRecordsLoadingState$ = this.store.pipe(select(technicalRecordsLoadingState));
   }
 
-  private reduceLoadingStates$() {
+  private get reduceLoadingStates$() {
     return combineLatest([this.globalLoadingState$, this.testResultLoadingState$, this.techRecordsLoadingState$]).pipe(
       map((states) =>
```
danielchildsbjss, line 24: Can this be
states.some(s => s === true);

src/app/layout/spinner/spinner.service.ts
mcadenas-bjss: I added an action we can use to change global state but at the moment nothing is using that as we don't have any non API process that might need a spinner. Later if we have any reason to show the spinner that doesn't involve an API we can use that action.

src/app/layout/spinner/spinner.service.ts
mcadenas-bjss: ooh didn't think of that!

danielchildsbjss (approved):

kpearmanbjss (approved):

PR #334, opened 2022-05-27
fix: msal guard activation

gjulien-bjss:
## Fix MSAL Activation

_Change the msal activation bahviour back to original behaviour_
[link to ticket number]()

## Checklist


tom-bjss (changes requested):
Think only the Home Screen needs canActivate

andrewf-bjss (changes requested):
only the Home route needs this change

PR #344, opened 2022-06-13
feat(cb2-4030): Validators

gjulien-bjss:
## Validators

_Adding validators_
[CB2-4030](https://dvsa.atlassian.net/browse/CB2-4030)

## Checklist


src/app/forms/utils/error-message-map.ts
```diff
@@ -1,3 +1,5 @@
 export const ErrorMessageMap: { [key: string]: any } = {
-  ['required']: (err: boolean, label: string) => `${label ? label : 'This filed'} is required`
+  ['required']: (err: boolean, label: string) => `${label ? label : 'This filed'} is required`,
+  ['pattern']: (err: boolean, label: string) => `${label ? label : 'This filed'} must match a pattern`,
```
kpearmanbjss, line 3: Are these supposed to be "This field"?

gjulien-bjss: Yes it's the default in case the label doesn't exist

PR #349, opened 2022-06-15
Date component

mcadenas-bjss:
## Create Date Component

_Create a date component_
[link to ticket number](https://dvsa.atlassian.net/browse/CB2-4239)

## Checklist


src/app/forms/validators/date/date.validators.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+    [{ invalidDate: true }, '-01-00'],
+    [{ invalidDate: true }, ''],
+    [null, '2022-01-01'],
+    [null, '2022-01'],
+    [null, '2022']
+  ])('should validate date and return %s for %p', (expected, date) => {
+    const control = new FormControl(new Date(date));
+
+    expect(DateValidators.validDate(control)).toEqual(expected);
+  });
```
andrewf-bjss, line 19: Is it possible to have tests that prove the value can be read and stored as an ISO-8601 timestamp? (given that's how we've seen the data)

src/app/forms/utils/error-message-map.ts
```diff
@@ -1,3 +1,4 @@
 export const ErrorMessageMap: { [key: string]: any } = {
-  ['required']: (err: boolean, label: string) => `${label ? label : 'This filed'} is required`
+  required: (err: boolean, label: string) => `${label ? label : 'This filed'} is required`,
+  invalidDate: (err: boolean, label: string) => `${label ? label : 'This filed'} is an invalid date`
```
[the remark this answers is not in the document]
andrewf-bjss: Might be able to simplify the string interpolation with `${label ?? 'This field'} is required` too

src/app/forms/validators/date/date.validators.spec.ts
mcadenas-bjss: The value is propagated as Date object and should be converted to JSON string when we call the API. As for reading, the Date Pipe should handle the format as normal I think

gjulien-bjss (approved):

andrewf-bjss (approved):

PR #351, opened 2022-06-15
feature/CB2-3856: Searching for Multiple Tech Records with a Partial VIN

tahayazici:
## Searching for multiple Tech Records with a partial VIN

- Add a type selector to the search page.
- Make the errors dynamic on the search page.
- Add a component for displaying a single search result.
- Use that component to compose a list of those on a separate search results page.
- Handle the API call when searching by partial VIN.

[CB2-2891](https://dvsa.atlassian.net/browse/CB2-2891)

## Checklist


src/app/store/technical-records/effects/technical-record-service.effects.ts
```diff
@@ -5,55 +5,43 @@ import { of } from 'rxjs';
...
+      message = `There was a problem getting the Tech Record by ${search}`;
+    }
+  }
+  return message
+}
+
 @Injectable()
 export class TechnicalRecordServiceEffects {
-  getByVin$ = createEffect(() =>
+  getTechnicalRecord$ = createEffect(() =>
```
tahayazici, line 22: Loving the code reuse here 👍

src/app/features/search/multiple-search-results/multiple-search-results.component.ts
```diff
@@ -0,0 +1,46 @@
...
+  selector: 'app-multiple-search-results',
+  templateUrl: './multiple-search-results.component.html',
+  styleUrls: ['multiple-search-results.component.scss']
+})
+export class MultipleSearchResultsComponent implements OnDestroy {
+  vehicleTechRecords$: Observable<VehicleTechRecordModel[]>;
+  ngDestroy$ = new Subject();
+
+  constructor(public globalErrorService: GlobalErrorService, private technicalRecordService: TechnicalRecordService, private store: Store, private location: Location) {
+    this.store //
```
premghinde, line 20: // ?

src/app/features/search/multiple-search-results/multiple-search-results.component.ts
```diff
@@ -0,0 +1,46 @@
...
+            this.globalErrorService.clearError();
+            this.technicalRecordService.searchBy({ type, searchTerm });
+          }
+        }
+      });
+
+    this.vehicleTechRecords$ = this.technicalRecordService.vehicleTechRecords$;
+  }
+
+  navigateBack() {
```
premghinde, line 37: Maybe a reusable function?

tahayazici: What do you have in mind? My thought is that, in the near future when multiple pages require a back button, we could create a base component in the core module which all these pages would extend.

But this should be its own ticket and PR.

src/app/features/search/multiple-search-results/multiple-search-results.component.ts
tahayazici: I have Prettier messing up my code on saving and I put this to disable that. It's sorted now. This is removed.

src/app/features/search/multiple-search-results/multiple-search-results.component.ts
kpearmanbjss: Not sure if relevant but there was something on our side where we were deliberately not putting back buttons on pages, unsure if this is something that we're just doing by accident or if there's a reason for it?

src/app/forms/components/dynamic-form-group/dynamic-form-group.component.scss
```diff
@@ -8,3 +8,9 @@
     margin-top: 30px;
   }
 }
+
+.subheading-container {
```
gjulien-bjss, line 12: Could this be done with gov-uk classes ? https://design-system.service.gov.uk/styles/layout/

tahayazici: The gov-uk classes use set percentages such as full, one-half, one-third, two-thirds, one-quarter, three-quarters, etc.

Whereas, I want my 2 elements to have the minimum width possible and fill the middle with space, regardless of screen size.

naathanbrown (approved):

src/app/features/search/multiple-search-results/multiple-search-results.component.ts
tahayazici: I am resolving this due to inactivity, but do take it with me offline if you want to create some base shared components and functions as I'm all about that.

src/app/services/technical-record/technical-record.service.ts
```diff
@@ -25,8 +39,58 @@ export class TechnicalRecordService {
...
+        break;
+    }
+  }
+
+  /**
+   * A function to get the correct tech record to create the summary display which uses time first then status code
+   * @param record This is a VehicleTechRecordModel passed in from the parent component
+   * @returns returns the tech record of correct hierarchy precedence or if none exists returns undefined
+   */
+  viewableTechRecord(record: VehicleTechRecordModel, destroy$: Subject<any>): TechRecordModel | undefined {
```
mcadenas-bjss, line 60: Couple of observations here...
1. Could this be a completely asynchronous function so there is no race condition between the if on line 71 and the subscription. You could map the params and try to find the tech record, if not found resolve to the status code filter.
2. could al of this be done as a selector since all of this data should be in state?

tahayazici: This has now been addressed, thanks for the suggestion.

I will try and turn it into a selector in one of the next PRs as there's a couple more to finish off the search functionality.

naathanbrown (approved):

mcadenas-bjss (approved):

PR #353, opened 2022-06-20
feat(cb2-3994): amend vehicle section

gjulien-bjss:
## Amend vehicle section

_Ability to change the data inside of vehicle section_
[CB2-3994](https://dvsa.atlassian.net/browse/CB2-3994)

## Checklist

- [ ] Branch is rebased against the latest develop/common
- [ ] Necessary `id` required prepended with `"test-"` have been checked with automation testers and added
- [ ] Code and UI has been tested manually after the additional changes
- [ ] PR title includes the JIRA ticket number
- [ ] Squashed commits contain the JIRA ticket number
- [ ] Link to the PR added to the repo
- [ ] Delete branch after merge

PR #359, opened 2022-06-21
feat(cb2-3998): amend notes section

gjulien-bjss:
## Amend Notes Section

_Ability to edit notes section_
[CB2-3998](https://dvsa.atlassian.net/browse/CB2-3998)
depends on #353

## Checklist


mcadenas-bjss (dismissed):

mcadenas-bjss (approved):

src/app/forms/services/dynamic-form.service.ts
```diff
@@ -78,7 +79,7 @@ export class DynamicFormService {
     return controls;
   }
 
-  addValidators(control: CustomFormGroup | CustomFormArray | CustomFormControl, validators: Array<{ name: string; args?: any[] }> = []) {
+  addValidators(control: CustomFormGroup | CustomFormArray | CustomFormControl, validators: Array<{ name: string; args?: any }> = []) {
```
aaron-downing, line 82: If in the above args is `string` or `number`, could we not change this `any` here to `string | number`

gjulien-bjss: The issue with replacing the `any` to `string [] | number` (and I assume you missed the `string` on line 16&17) is that these are arguments into the validator functions, meaning we could theoretically have "`any`" input into the validatorMap and the validator functions.

In practice, because the binding from the template to the validator function is done through `addValidators` and `validatorMap`, and the template must ideally remain valid json, the input types are likely to grow to be `string | string[] | number | number[] | boolean | boolean[]` and maybe even `| object | object[]`. This, I believe, essentially amounts to having `any`.

Strong typing for validators is also hard to enforce through the template if it is to remain valid json as we'll only be able to use the validator `name` as a string to select which validator to apply to the control, meaning the arguments passed to the validator in the form could be mis-matched with the validator type. This is unavoidable if the template is to remain valid json. We decided the drawback of not being able to have strong typing for passing arguments to validator functions from the template is outweighed by the benefit of keeping the template as json.

PR #393, opened 2022-07-06
fix(cb2-4566): fix tech record not rendering with an empty array

naathanbrown:
## Fix tech record not rendering with an empty array

When a section that requires an array to be provided was given a null, the form broke and would not load. This fix aims to correct that. Currently it just renders the title and no more information, but in future work this will be changed.

## Checklist

### DIFF
--- a/src/app/features/tech-record/components/tech-record-summary/tech-record-summary.component.ts
+++ b/src/app/features/tech-record/components/tech-record-summary/tech-record-summary.component.ts
@@ -25,7 +25,7 @@ import { getDimensionsMinMaxSection, getDimensionsSection } from '@forms/templat
 import { getBodyTemplate } from '@forms/templates/general/body.template';
 import { NotesTemplate } from '@forms/templates/general/notes.template';
 import { DocumentsTemplate } from '@forms/templates/general/documents.template';
-import { PlatesTemplate } from '@forms/templates/general/plates.template'
+import { PlatesTemplate } from '@forms/templates/general/plates.template';
 import { TrlAuthIntoServiceTemplate } from '@forms/templates/trl/trl-auth-into-service.template';
 import { TrlManufacturerTemplate } from '@forms/templates/trl/trl-manufacturer.template';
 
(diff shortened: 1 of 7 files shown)


src/app/features/tech-record/components/tech-record-summary/tech-record-summary.component.ts
```diff
@@ -76,7 +76,11 @@ export class TechRecordSummaryComponent implements OnInit {
         this.psvBrakeTemplate = PsvBrakeSection;
         this.brakeTemplateWheelsNotLocked = PsvBrakeSectionWheelsNotLocked;
         this.brakeTemplateWheelsHalfLocked = PsvBrakeSectionWheelsHalfLocked;
-        this.dimensionsTemplate = getDimensionsSection(VehicleTypes.PSV, this.vehicleTechRecord.noOfAxles, this.vehicleTechRecord?.dimensions?.axleSpacing);
+        this.dimensionsTemplate = getDimensionsSection(
+          VehicleTypes.PSV,
+          this.vehicleTechRecord.noOfAxles,
+          this.vehicleTechRecord?.dimensions?.axleSpacing
```
Ian-Hodges, line 82: (I realise this is not your change) Is there a chance vehicleTechRecord could be null? The optional chaining operator is used inconsistently on these two lines and further below.

naathanbrown: A solid point, it can be null as far as I know, I will go an add these in to all of them.

gjulien-bjss (approved):

src/app/forms/services/dynamic-form.service.ts
```diff
@@ -62,13 +62,13 @@ export class DynamicFormService {
 
   createControls(child: FormNode, d: any) {
     const controls: any[] = [];
-    if (d.length && d.length > 0) {
+    if (Array.isArray(d)) {
```
tahayazici, line 65: This is a clever check because if is `d` is an `Array`, it won't matter if it's empty or not. Nice 👌

tahayazici (approved):

PR #403, opened 2022-07-13
feat(cb2-4682): add required hidden fields to templates

gjulien-bjss:
## Add required hidden fields to templates

_Adding required hidden fields to the templates_
- Fix typo in develop
- Downgrading `package-lock.json`, `npm audit` fails with `node 14`, meaning jenkins builds will fail

Part of [CB2-4642](https://dvsa.atlassian.net/browse/CB2-4642)

## Checklist


src/app/forms/templates/test-records/section-templates/test/test-section.template.ts
```diff
@@ -123,6 +123,17 @@ export const TestSection: FormNode = {
               disabled: true,
               label: 'End time',
               viewType: FormNodeViewTypes.TIME
+            },
+            {
+              name: 'prohibitionIssued',
+              type: FormNodeTypes.CONTROL,
+              value: '',
+              label: 'Prohibition issued',
+              options: [
```
tahayazici, line 132: Do you not need to specify the view and edit type for this control?

src/app/forms/templates/test-records/section-templates/test/test-section-group1.template.ts
```diff
@@ -0,0 +1,132 @@
+import { FormNode, FormNodeEditTypes, FormNodeTypes, FormNodeViewTypes } from '@forms/services/dynamic-form.types';
+
+export const TestSectionGroup1: FormNode = {
```
tahayazici, line 3: Why is this called `TestSectionGroup1`? I do not see any group 2 or 3 🙂

src/app/forms/templates/test-records/section-templates/test/test-section.template.ts
gjulien-bjss: You don't have to because I just want the default viewType and we're not handling edit yet

src/app/forms/templates/test-records/section-templates/test/test-section-group1.template.ts
gjulien-bjss: It's to do with the taxonomy that we know we're going to have to handle in the future, we'll probably have to handle around 16 groups, this is just the first step, it'll have to be refactored later

mcadenas-bjss (approved):

src/app/features/test-records/components/base-test-record/base-test-record.component.ts
```diff
@@ -19,11 +19,12 @@ export class BaseTestRecordComponent {
 
   generateTemplate(): FormNode[] | undefined {
```
tahayazici, line 20: This is a bit of a nesting-ception and `this.testResult.testTypes[0].testTypeId` is being accessed without checking if the array has at least one item. Also, the initial `if (this.testResult)` check is redundant as this component requires that input to function, so you can enforce that outside of this method and then flatten this method a lil bit.

Enforce the input in the selector:
```TypeScript
@Component({
  selector: 'app-base-test-record[testResult]',
  templateUrl: './base-test-record.component.html'
})
```
Confirm that this will be populated:
```TypeScript
  @Input() testResult!: TestResultModel;
```
Clean up `generateTemplate()`:
```TypeScript
  generateTemplate(): FormNode[] | undefined {
    const { vehicleType } = this.testResult;

    if (!vehicleType || !masterTpl.hasOwnProperty(vehicleType) || this.testResult.testTypes.length == 0) {
      return undefined;
    }

    const testTypeId = this.testResult.testTypes[0].testTypeId;
    const vehicleTpl: Partial<Record<string | 'default', Record<string, FormNode>>> = masterTpl[vehicleType as VehicleTypes];

    const tpl = testTypeId && vehicleTpl.hasOwnProperty(testTypeId)
      ? vehicleTpl[testTypeId]
      : vehicleTpl['default'];
    
    return tpl && Object.values(tpl);
  }
```

tahayazici: Damn... I'm 2 mins too late. Never mind, we can address this later on in another PR.

PR #424, opened 2022-07-21
refactor(*): refactor tech record routing

gjulien-bjss:
## Refactor tech record routing

_Amongst other things this PR aims the refactor the technical record routing and a few cleanups_
- Trigger the loading spinner when in tech record view if the test records have not been loaded yet
- Introduce a guard on the search results page to redirect to the home page (if signing off and on the results page, the page resolved to a blank page)
- Use a resolver for technical record data
- Use `systemNumber` in the tech record routing as it is a unique value, unlike `vin`
- Added functionality to search by all criteria, which is an existing endpoint, and made default option in the select
## Checklist


src/app/features/tech-record/tech-record-routing.module.ts
```diff
@@ -1,18 +1,21 @@
...
-    canActivateChild: [MsalGuard]
+    canActivateChild: [MsalGuard],
+    resolve: { load: TechRecordViewResolver }
   },
   {
-    path: ':vin/:techCreatedAt',
+    path: ':systemNumber/:techCreatedAt',
     component: TechRecordComponent,
-    data: { title: 'Tech Record' }
+    data: { title: 'Tech Record' },
```
aaron-downing, line 17: Do we need the canActivateChild guard here too?

aaron-downing (approved):

src/app/features/tech-record/tech-record-routing.module.ts
naathanbrown: I am not sure it is needed on either. This route does not have a child route, I think it is redundant code.

naathanbrown (approved):

PR #425, opened 2022-07-21
feat(cb2-4690): create frontend auth

naathanbrown:
## Enable front end role based access

Create the guard and directive for front end auth.
[4960](https://dvsa.atlassian.net/browse/CB2-4724)

## Checklist


src/app/app-routing.module.ts
```diff
@@ -10,14 +11,14 @@ const routes: Routes = [
     children: [
       {
         path: '',
-        data: { title: 'Home' },
-        canActivate: [MsalGuard],
+        data: { title: 'Home', roles: ['CVSFullAccess', 'TechRecord.Read'] },
```
aaron-downing, line 14: This'll be TechRecord.View not read

src/app/directives/user-role.directive.ts
```diff
@@ -0,0 +1,34 @@
...
+@Directive({ selector: '[appUserRole]' })
+export class UserRoleDirective implements OnInit {
+  constructor(private templateRef: TemplateRef<any>, private userService: UserService, private viewContainer: ViewContainerRef) {}
+
+  userRoles: string[] | undefined;
+
+  @Input()
+  set appUserRole(roles: string[]) {
+    if (!roles || !roles.length) {
+      throw new Error('Roles value is empty or missed');
```
andrewf-bjss, line 14: Nitpick...

```suggestion
      throw new Error('Roles value is empty or missing');
```

src/app/app-routing.module.ts
andrewf-bjss, line 14: Could we have `VTMAdmin` added to this list wherever `CVSFullAccess` is mentioned as well please?

andrewf-bjss (changes requested):

src/app/app-routing.module.ts
naathanbrown: Thanks, this is honestly just placeholder for now as none of it even exits in AD yet haha.

naathanbrown: Hi, yes, I have currently left this off so I can test this before platform actually do there bit. When this moves to develop, I will add in this everywhere too, but leaving it off for now allows me to use a CVSFullAccess to see it works, then use a VTMAdmin to prove it does not

tom-bjss (approved):

andrewf-bjss:
`package-lock.json` - please recreate this with node v14 so it works with `npm audit`  (lockfile version should be 1)

src/app/app-routing.module.ts
andrewf-bjss: OK, but if this is merged into `develop` then most if not all of the devs will be locked out!

Could we merge this into a feature branch instead, verify the change there, then add the `VTMAdmin` role before merging into `develop` please?

andrewf-bjss: Sorry - I see you are intending to merge this into `VTMDEV-1` feature env, thank you!

andrewf-bjss (approved):

src/app/features/search/multiple-search-results/multiple-search-results.component.html
```diff
@@ -4,6 +4,10 @@
   </div>
 
   <ng-container *ngFor="let vehicleTechRecord of vehicleTechRecords$ | async">
-    <app-single-search-result class="govuk-grid-column-full" [vehicleTechRecord]="vehicleTechRecord"></app-single-search-result>
+    <app-single-search-result
+      class="govuk-grid-column-full"
+      [vehicleTechRecord]="vehicleTechRecord"
+      *appUserRole="['CVSFullAccess', 'TechRecord.View']"
```
mcadenas-bjss, line 10: Wouldn't this view be protected by the guard rather than hiding content which might be a bad user experience as they would think the view is broken?

src/app/features/search/search.component.html
```diff
@@ -1,4 +1,4 @@
-<div class="govuk-grid-row search-page">
+<div class="govuk-grid-row search-page" *appUserRole="['CVSFullAccess', 'TechRecord.View']">
```
mcadenas-bjss, line 1: Same as before. rather than letting users in here and hiding content, would it not be better to guard the route?

src/app/guards/roles.guard.ts
```diff
@@ -0,0 +1,21 @@
...
+  constructor(private userService: UserService) {}
+
+  canActivate(next: ActivatedRouteSnapshot): boolean {
+    let matchingRoles = false;
+    this.userService.roles$.pipe(take(1)).subscribe(storedRoles => {
+      const allowedRoles: string[] = next.data['roles'];
+      matchingRoles = allowedRoles.some(x => storedRoles.includes(x));
+    });
+
+    return matchingRoles;
```
mcadenas-bjss: What happens when we reject the navigation? Are we stuck on some kind of limbo or should be navigate to Home screen?

src/app/features/search/multiple-search-results/multiple-search-results.component.html
naathanbrown: The route is guarded too, I took a both approch. It maybe worth considering if this is needed, especially as the API is never hit as it is protected too, but why would it ever be a bad UX? If you have the permission you will see it, if you don't then it doesn't matter. You need to be able to view a tech record to see it, this in my view is fine to hide it behind this.

src/app/guards/roles.guard.ts
naathanbrown: You simply get shown a blank screen, like just the header appears, which is the desired effect.

PR #433, opened 2022-08-01
VTM Role Based Access

tom-bjss:
This PR merges VTMDEV-1 into develop to add role based access. Ticket has been tested and evidence attached here: https://dvsa.atlassian.net/browse/CB2-4724


mcadenas-bjss (commented):
Is it necessary to use the directive on components that are already protected by the guard?

mcadenas-bjss (approved):

Sul-Aga-kainos:
> > Is it necessary to use the directive on components that are already protected by the guard?
> 
> Not 'necessary' but I had a conversation with Steven Pears about this and we agreed it's a net positive. The code is not that invasive and more layers of protection is never a bad thing.

I think security measurements (using guards or in place using directive) should be used as required and not mixed because this will force you to write unit tests that you don't actually need.

PR #449, opened 2022-08-22
feat(cb2-4611): amend test result

gjulien-bjss:
## Automation of Test Result amendment (Test Section)

_This story covers review and confirmation of test result auto amendments as per defects updates_
[CB2-4611](https://dvsa.atlassian.net/browse/CB2-4611)

## Checklist


mcadenas-bjss (dismissed):

kpearmanbjss (dismissed):

src/app/forms/components/defect/defect.component.html
```diff
@@ -83,8 +83,17 @@
       <app-text-input formControlName="notes" label="Notes"></app-text-input>
     </ng-container>
 
-    <app-text-input formControlName="prs" label="PRS"></app-text-input>
     <app-radio-group
+      name="prs"
+      formControlName="prs"
+      label="PRS"
+      [options]="[
```
tahayazici, line 90: This is a very chuncky object to go into the HTML, maybe consider having a getter in the component for it?
```typescript
get options(): Option[] {
  return [
    { value: true, label: 'Yes' },
    { value: false, label: 'No' }
  ];
}
```

src/app/forms/templates/test-records/section-templates/test/test-section-group1.template.ts
```diff
@@ -47,13 +47,8 @@ export const TestSectionGroup1: FormNode = {
             {
               name: 'testResult',
               label: 'Result',
-              editType: FormNodeEditTypes.RADIO,
-              options: [
-                { value: 'fail', label: 'Fail' },
-                { value: 'pass', label: 'Pass' },
-                { value: 'prs', label: 'PRS' },
-                { value: 'abandoned', label: 'Abandoned' }
-              ],
```
Ruairidh-BJSS, line 56: By hiding the editType how does the user select to abandon the test?

src/app/services/result-of-test/result-of-test.service.ts
```diff
@@ -0,0 +1,21 @@
...
+import { resultOfTestEnum } from '@models/test-types/test-type.model';
+import { select, Store } from '@ngrx/store';
+import { State } from '@store/.';
+import { resultOfTestSelector, updateResultOfTest } from '@store/test-records';
+import { Observable } from 'rxjs';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class ResultOfTestService {
```
tahayazici, line 11: This adheres to our current pattern and the decision to use this was made by someone else who is not in the project anymore, so it is fine. But just an observation here:

Both methods here are one-liners. The only thing this service achieves is to hide the `store`. In other words, abstraction for abstraction's sake. In object-oriented programming, abstraction is used to reduce complexity and increase efficiency. I'll use one example...

```typescript
  calculateResult(sequence: number[], exponent: number): number {
    const integral = this.mathsService.evaluateIntegralSequence(sequence); // contains several loops

    const fibonacci = this.mathsService.getNearestFibonacci(exponent); // works with a huge switch case

    return this.mathsService.recurseAndCombine(integral, fibonacci); // consists of several recursive methods
  }
```
Here the method is concerned with applying several steps and returning their result. The internals of those steps is long and complex. Without abstraction, this method could be 30 lines and its flow would not be apparent anymore.


---
```typescript
  getSum(a: number, b: number): number {
    return this.mathsService.addTwoNumbers(a, b);
  }

  class mathsService {
    addTwoNumbers(a: number, b: number): number {
      return a + b;
    }
  }
```
Whilst here, we are moving just one line of code from one place to another.

I think it would be a fruitful conversation to decide on this sometime in the future (before it's too late) @mcadenas-bjss  & @andrewf-bjss. And as a result, if the consensus is to still abstract with services, then I'm happy to comply - but at least we would have made that decision explicitly.

src/app/store/test-records/selectors/test-records.selectors.ts
```diff
@@ -82,9 +83,12 @@ export const selectAmendedDefectData = createSelector(selectedAmendedTestResultS
...
 export const sectionTemplates = createSelector(testResultsFeatureState, state => state.sectionTemplates);
 
-export const isSameTestTypeId = createSelector(selectedAmendedTestResultState, selectedTestResultState, (testRecord, amendedTestRecord) => {
-  return testRecord?.testTypes[0].testTypeId === amendedTestRecord?.testTypes[0].testTypeId;
-});
+export const resultOfTestSelector = createSelector(toEditOrNotToEdit, testRecord => testRecord?.testTypes[0].testResult);
+
+export const isTestTypeKeySame = (key: keyof TestType) =>
+  createSelector(selectedAmendedTestResultState, selectedTestResultState, (testRecord, amendedTestRecord) => {
+    return testRecord?.testTypes[0][key] === amendedTestRecord?.testTypes[0][key];
```
tahayazici, line 90: Is it guaranteed that `testRecord` and `amendedTestRecord` will always have at least one `testTypes`?

src/app/forms/templates/test-records/section-templates/test/test-section-group1.template.ts
gjulien-bjss: We don't need to cater for that scenario, abandoned tests will stay abandoned and there is no requirement to set tests as abandoned VTM

src/app/store/test-records/selectors/test-records.selectors.ts
gjulien-bjss: They would both have at least one test type is they are defined

mcadenas-bjss (approved):

src/app/forms/components/defect/defect.component.html
tahayazici: Alright, thanks for trying.

PR #459, opened 2022-08-25
fix(cb2-5499): update test station details

gjulien-bjss:
## Test station type does not update on selection

_Test station type does not update on selection_
[CB2-5499](https://dvsa.atlassian.net/browse/CB2-5499)

## Checklist

### DIFF
--- a/src/app/forms/validators/custom-async-validators.ts
+++ b/src/app/forms/validators/custom-async-validators.ts
@@ -34,7 +34,6 @@ export class CustomAsyncValidators {
       store
         .pipe(select(getTestStationFromProperty((control as CustomFormControl).meta.name as keyof TestStation, control.value)), take(1))
         .subscribe(stations => {
-          // store.dispatch(updateResultOfTest());
           stations && store.dispatch(updateTestStation({ payload: stations }));
         });
       return of(null);
--- a/src/app/forms/validators/custom-async-validators.ts
+++ b/src/app/forms/validators/custom-async-validators.ts
@@ -1,10 +1,14 @@
 import { AbstractControl, AsyncValidatorFn, ValidationErrors } from '@angular/forms';
+import { CustomFormControl } from '@forms/services/dynamic-form.types';
+import { TestStation } from '@models/test-stations/test-station.model';
 import { select, Store } from '@ngrx/store';
-import { testResultInEdit, TestResultsState } from '@store/test-records';
-import { map, Observable, take } from 'rxjs';
+import { State } from '@store/.';
+import { testResultInEdit } from '@store/test-records';
+import { getTestStationFromProperty, updateTestStation } from '@store/test-stations';
+import { map, Observable, of, take } from 'rxjs';
 
 export class CustomAsyncValidators {
-  static resultDependantOnCustomDefects(store: Store<TestResultsState>): AsyncValidatorFn {
+  static resultDependantOnCustomDefects(store: Store<State>): AsyncValidatorFn {
     return (control: AbstractControl): Observable<ValidationErrors | null> =>
       store.pipe(
         take(1),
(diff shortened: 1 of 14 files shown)


Ruairidh-BJSS (dismissed):

CormacBJSS (approved):

src/app/store/test-records/reducers/test-records.reducer.ts
```diff
@@ -65,7 +66,20 @@ export const testResultsReducer = createReducer(
...
+    return {
+      ...state,
+      editingTestResult: {
+        ...state.editingTestResult,
+        testStationName: action.payload.testStationName,
+        testStationType: action.payload.testStationType
+      }
+    };
+  })
 );
```
tahayazici, line 83: Just to make it a bit more concise
```suggestion
  on(updateTestStation, (state, action) => !state.editingTestResult
    ?  { ...state }
    : {
      ...state,
      editingTestResult: {
        ...state.editingTestResult,
        testStationName: action.payload.testStationName,
        testStationType: action.payload.testStationType
      }
    }
);
```

src/app/store/test-stations/actions/test-stations.actions.ts
```diff
@@ -15,3 +15,5 @@ function getTitle(isPlural: boolean = false, suffix: string = ''): string {
   suffix = suffix ? ' ' + suffix : suffix;
   return '[API/test-stations] Fetch Test Station' + plural + suffix;
 }
+
+export const updateTestStation = createAction('[test-stations] update the test station', props<{ payload: TestStation }>());
```
tahayazici, line 19: You could extend `getTitle()` to take in an `action` parameter and use that.

src/app/forms/validators/custom-async-validators.ts
```diff
@@ -24,4 +28,15 @@ export class CustomAsyncValidators {
...
       );
   }
+
+  static updateTestStationDetails(store: Store<State>): AsyncValidatorFn {
+    return (control: AbstractControl): Observable<null> => {
+      store
+        .pipe(select(getTestStationFromProperty((control as CustomFormControl).meta.name as keyof TestStation, control.value)), take(1))
+        .subscribe(stations => {
+          stations && store.dispatch(updateTestStation({ payload: stations }));
+        });
```
tahayazici, line 38: You were thinking of where to use `tap()`, right? 🙂
```suggestion
      const propertyName = (control as CustomFormControl).meta.name as keyof TestStation;
      
      store
        .pipe(
          select(getTestStationFromProperty(propertyName, control.value)),
          take(1),
          tap(stations => stations && store.dispatch(updateTestStation({ payload: stations })))
        )
        .subscribe();
```

src/app/store/test-records/reducers/test-records.reducer.ts
gjulien-bjss: Nice

src/app/forms/validators/custom-async-validators.ts
gjulien-bjss: I did think about using the tap, but decided against it and just put it in the subscribe instead

src/app/store/test-stations/actions/test-stations.actions.ts
tahayazici: That's fair.

src/app/forms/validators/custom-async-validators.ts
tahayazici: Don't you think it looks cleaner within `tap()`? It's like an effect of the store - do go and trigger something else after getting the value of the observable.

kpearmanbjss (approved):

src/app/forms/validators/custom-async-validators.ts
mcadenas-bjss: feels like that would be using an operator for the sake of it. You have to subscribe anyway so might as well use it.

mcadenas-bjss (approved):

PR #491, opened 2022-09-21
fix(cb2-4379): manually change test result

CormacBJSS:
## Ticket title
Allow test result to be manually changed for test types without a defects section
[https://dvsa.atlassian.net/browse/CB2-4379]

## Checklist


src/app/forms/validators/custom-async-validators.ts
```diff
@@ -82,4 +82,31 @@ export class CustomAsyncValidators {
       );
     };
   }
-}
+
+  static testAndSwitchToHiddenFieldWithDefectTaxonomy(store: Store<State>): AsyncValidatorFn {
```
gjulien-bjss, line 86: Would it not be better to have this templated? If tests use defects taxonomy, then just make the result hidden, if they don't make it radiobuttons. Seems like this implementation somewhat defeats the purpose of templates

CormacBJSS: That has been done - the async validators have been removed and the group test section templates have been changed to specify if the test result field is hidden or a radio button.

mcadenas-bjss (approved):

kpearmanbjss (approved):

PR #502, opened 2022-09-26
feat(cb2-5728): auto focus date segments

mcadenas-bjss:
## Date Component Improvement

_Automate the focus event between date segments as the user types numeric values in._
[link to ticket number]()

## Checklist

### DIFF
--- a/src/app/forms/components/date/date.component.html
+++ b/src/app/forms/components/date/date.component.html
@@ -14,7 +14,7 @@
             [includeTime]="includeTime"
             class="govuk-input govuk-date-input__input govuk-input--width-2"
             [ngClass]="{ 'govuk-input--error': elementHasErrors(0) && (dayEl.dirty || dayEl.touched) }"
-            attr.id="{{ name }}-day"
+            id="{{ name }}-day"
             name="{{ name }}-day"
             type="number"
             inputmode="numeric"
--- a/src/app/forms/components/date/date.component.html
+++ b/src/app/forms/components/date/date.component.html
@@ -10,9 +10,11 @@
           <input
             #dayEl="ngModel"
             appNumberOnly
+            appFocusNext
+            [includeTime]="includeTime"
             class="govuk-input govuk-date-input__input govuk-input--width-2"
             [ngClass]="{ 'govuk-input--error': elementHasErrors(0) && (dayEl.dirty || dayEl.touched) }"
-            id="{{ name }}-day"
+            attr.id="{{ name }}-day"
             name="{{ name }}-day"
             type="number"
             inputmode="numeric"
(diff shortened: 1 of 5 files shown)


src/app/forms/components/date/focus-next.directive.ts
```diff
@@ -0,0 +1,40 @@
...
+      }
+    } else if (value.length === 4 && this.includeTime && 'year' === currentSegment) {
+      this.focus(`${name}-hour`);
+      return;
+    }
+  }
+
+  focus(el: string) {
+    document.getElementById(el)?.focus();
+  }
```
tahayazici, line 39: As requested, here is an alternative in which calculating the suffix is abstracted:
```suggestion
    const segments = id.split('-');

    const suffix = this.getSuffix(value, segments.splice(-1)[0]);

    if (suffix) {
      document.getElementById(segments.join('-') + suffix)?.focus();
    }
  }

  getSuffix(value: any, currentSegment: any): string | undefined {
    if (value.length === 2) {
      if ('day' === currentSegment) {
        return '-month';
      } else if ('month' === currentSegment) {
        return '-year';
      } else if ('hour' === currentSegment) {
        return '-minute';
      }
    } else if (value.length === 4 && this.includeTime && currentSegment === 'year') {
      return '-hour';
    }
  }
```

tahayazici, line 39: As requested, here is an alternative in which calculating the suffix is abstracted:
```suggestion
  @HostListener('input', ['$event'])
  onInput(e: InputEvent) {
    const segments = this.el.nativeElement.id.split('-');

    const suffix = this.getSuffix(segments.splice(-1)[0]);

    if (suffix) {
      document.getElementById(segments.join('-') + suffix)?.focus();
    }
  }

  getSuffix(currentSegment: any): string | undefined {
    const { nativeElement: value } = this.el;

    if (value.length === 2) {
      if ('day' === currentSegment) {
        return '-month';
      } else if ('month' === currentSegment) {
        return '-year';
      } else if ('hour' === currentSegment) {
        return '-minute';
      }
    } else if (value.length === 4 && this.includeTime && currentSegment === 'year') {
      return '-hour';
    }
  }
```

mcadenas-bjss: Thank, good idea! I've implemented it with some small changes.

src/app/forms/components/date/date.component.html
```diff
@@ -10,9 +10,11 @@
           <input
             #dayEl="ngModel"
             appNumberOnly
+            appFocusNext
+            [includeTime]="includeTime"
             class="govuk-input govuk-date-input__input govuk-input--width-2"
             [ngClass]="{ 'govuk-input--error': elementHasErrors(0) && (dayEl.dirty || dayEl.touched) }"
-            id="{{ name }}-day"
+            attr.id="{{ name }}-day"
```
gjulien-bjss, line 17: Why the change? especially since the other components don't have that change?
```suggestion
            id="{{ name }}-day"
```

gjulien-bjss (approved):

kpearmanbjss (approved):

PR #507, opened 2022-09-27
fix(*): handle unknown routes

mcadenas-bjss:
## Page no found

_Handle unknown routes with redirects and "Page not found" view._

## Checklist

### DIFF
--- a/src/app/features/tech-record/tech-record-routing.module.ts
+++ b/src/app/features/tech-record/tech-record-routing.module.ts
@@ -1,10 +1,10 @@
 import { NgModule } from '@angular/core';
 import { RouterModule, Routes } from '@angular/router';
 import { MsalGuard } from '@azure/msal-angular';
-import { TechRecordViewResolver } from 'src/app/resolvers/tech-record-view/tech-record-view.resolver';
-import { TechRecordComponent } from './tech-record.component';
 import { RoleGuard } from '@guards/roles.guard';
 import { Roles } from '@models/roles.enum';
+import { TechRecordViewResolver } from 'src/app/resolvers/tech-record-view/tech-record-view.resolver';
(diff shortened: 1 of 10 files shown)


src/app/features/tech-record/tech-record-routing.module.ts
```diff
@@ -15,7 +15,7 @@ const routes: Routes = [
     resolve: { load: TechRecordViewResolver }
   },
   {
-    path: ':techCreatedAt',
+    path: 'historic/:techCreatedAt',
```
aaron-downing, line 18: Probably out of the scope of this PR but now that the current record is the default record to be presented on `/tech-records/{systemNumber}`, if you route to provisional it would now be called 'historic' as well as if you went from the provisional back to the current. 

Initial:
image
After going to provisional:
image
Back to current:
image

kpearmanbjss (dismissed):

tahayazici (approved):

src/app/features/tech-record/tech-record-routing.module.ts
mcadenas-bjss: This change is so the 'historic' route can be differentiated from an invalid route. Without this change, `tech-records/3506666/fhjkdsgfjhdsf` would still resolve to the historic view while we actually want it to be 'Page not found'. Having an extra segment like `historic/:createdAt` will give us better control of the routing.
The issue you described above is part of my explanation why the historic tech record view should be the end of the line [here](https://dvsa.atlassian.net/browse/CB2-5982). Only the current record should show the history table and test records. Particularly when test record are on a system number basis, we shouldn't give the impression of showing tests for a 'historic' version.

tahayazici (approved):

naathanbrown (approved):

PR #510, opened 2022-09-27
fix(cb2-6113): handle duplicate system numbers in tech record db

aaron-downing:
## Bug: Unable to select vehicle with duplicate system number in multiple search result

On some occasions, it is possible for there to be more than one vehicles with the same `systemNumber` but different `vin`. 

This led to issues with the current VTM setup due to it being believed that `systemNumber` was a unique identifier for a vehicle and so navigating to a URL such as `http://localhost:4200/tech-records/{systemNumber}` could be the route to more than one vehicle.

The default route for a vehicle technical record has now been updated to `http://localhost:4200/tech-records/{systemNumber}/{vin}`.

[link to ticket number](https://dvsa.atlassian.net/browse/CB2-6113)

## Checklist


src/app/services/technical-record/technical-record.service.ts
```diff
@@ -7,7 +7,7 @@ import {
   editableTechRecord,
   getByAll,
   getByPartialVin,
-  getBySystemNumber,
+  getBySystemNumberAndVin,
```
gjulien-bjss, line 10: Is the import needed?

aaron-downing: will bin it

src/app/features/tech-record/components/tech-record-history/tech-record-history.component.html
```diff
@@ -19,12 +19,7 @@
       <td class="govuk-table__cell">{{ techRecord.createdAt | date: 'dd/MM/yyyy HH:mm' }}</td>
       <td class="govuk-table__cell">
         <ng-container *ngIf="techRecord.createdAt !== currentRecord?.createdAt">
-          <a
-            id="view-test-{{ vehicleTechRecord.vin }}"
-            href="/tech-records/{{ vehicleTechRecord.systemNumber }}/historic/{{ convertToUnix(techRecord.createdAt) }}"
-          >
-            View
-          </a>
+          <a id="view-test-{{ vehicleTechRecord.vin }}"> View </a>
```
gjulien-bjss, line 22: `href`/`routerlink` not needed anymore?

aaron-downing: Hmm that's interesting, the linter removed it...

src/app/forms/templates/search/single-search-result.template.ts
```diff
@@ -1,15 +1,15 @@
 import { ValidatorNames } from '@forms/models/validators.enum';
 import { FormNode, FormNodeTypes, FormNodeViewTypes } from '../../services/dynamic-form.types';
 
-export function createSingleSearchResult(systemNumber?: string): FormNode {
+export function createSingleSearchResult(systemNumber?: string, vin?: string): FormNode {
```
tahayazici, line 4: Should the parameters `systemNumber` and `vin` not be mandatory here?

If a call would be made to this function with both set to `null`, the URL would be "/tech-records//" and that would crash, right?

tahayazici (commented):
Good stuff, just one comment to bring clarity to the `createSingleSearchResult()` function.

src/app/forms/templates/search/single-search-result.template.ts
aaron-downing: This is already an issue if you don't pass it a systemNumber which then reroutes you back to the home page.

kpearmanbjss (approved):

tahayazici (approved):

PR #513, opened 2022-09-29
feat/cb2-5808-contingency-group-10

kpearmanbjss:
Contingency tests for group 10

Altered the anniversary date so it can't be edited as it is automatically generated on the backend.
[link to ticket number](https://dvsa.atlassian.net/browse/CB2-5808)

## Checklist


andrewf-bjss (approved):

CormacBJSS (approved):

src/app/forms/templates/test-records/section-templates/test/test-section-group1.template.ts
```diff
@@ -120,7 +120,7 @@ export const TestSectionGroup1: FormNode = {
               disabled: true,
               label: 'Anniversary date',
               viewType: FormNodeViewTypes.DATE,
-              editType: FormNodeEditTypes.DATE
+              editType: FormNodeEditTypes.HIDDEN
```
gjulien-bjss, line 123: Do we not want to still display these even though in amend mode? I understand not showing the field for contigency, not sure about hiding it in amend, especially if it's visible in view mode

kpearmanbjss: My point is that it's a field that shouldn't be editable because it's allegedly updated on the back end and that it should be visible in view mode as it's data related to the test record but the user shouldn't be able to manually change it.

tom-bjss: Heres my thoughts:

1. if its not editable but the value here would benefit editing others, it should still be viewable on the screen

2. The other thing is, with this missing would it confuse the user / make them think the record has missing data ? and result in a false negative bug being reported ?

Just my thoughts

kpearmanbjss: I guess those are reasons we could have it displayed but still marked as disabled, just wary of us having disabled fields viewable on the edit form so far. I'll change it back for the amend fields for now then.

PR #519, opened 2022-09-30
CB2-5686: Amend documents section for technical records

tom-bjss:
Amend documents for all technical records.

Added the service design for dropdown to have the minimal arrow, this will be reflected on test teams also.

I do wonder about the interface FormNodeOptions takes, since the 2 examples ive found for it just duplicate the string twice, would it be better to just be a string array ?

### DIFF
--- a/src/app/accessible-autocomplete.d.ts
+++ b/src/app/accessible-autocomplete.d.ts
@@ -14,6 +14,7 @@ declare module 'accessible-autocomplete/dist/accessible-autocomplete.min' {
     defaultValue?: string;
     autoselect?: boolean;
     showAllValues?: boolean;
+    dropdownArrow?: Function;
   }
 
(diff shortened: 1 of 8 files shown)


gjulien-bjss (approved):

gjulien-bjss (approved):
Looks good, don't want to tackle doing the document types using the reference data endpoint just yet? 😛 
- `FormNodeOptions` does have it issues being in the `label - value` format as it is right now, I think it is definitely useful for something like country of registration which has a value such as `gb` and a label of `Great Britain` for example. We had some issues changing it to a union type such as 
```ts
options: string[] | {label: string, value: string | number}[]
```
because the compiler doesn't seem to be able to figure out what format the `options` field will have.

Could also be we won't have to tackle that problem when we come to using reference data

kpearmanbjss (approved):

PR #533, opened 2022-10-04
Amend dda section

Ian-Hodges:
Enable amendment of the DDA section.
[CB2-5689](https://dvsa.atlassian.net/browse/CB2-5689)

### DIFF
--- a/src/app/forms/templates/psv/psv-dda.template.ts
+++ b/src/app/forms/templates/psv/psv-dda.template.ts
@@ -1,4 +1,5 @@
-import { FormNode, FormNodeTypes } from '../../services/dynamic-form.types';
+import { ValidatorNames } from '@forms/models/validators.enum';
+import { FormNode, FormNodeEditTypes, FormNodeTypes, FormNodeViewTypes } from '../../services/dynamic-form.types';
 
 export const PsvDdaTemplate: FormNode = {
   name: 'dda',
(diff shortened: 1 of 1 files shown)


src/app/forms/templates/psv/psv-dda.template.ts
```diff
@@ -13,73 +14,139 @@ export const PsvDdaTemplate: FormNode = {
...
         },
         {
           name: 'wheelchairRampPresent',
           label: 'Wheelchair ramp present',
           value: '',
-          type: FormNodeTypes.CONTROL
+          type: FormNodeTypes.CONTROL,
+          viewType: FormNodeViewTypes.STRING,
+          editType: FormNodeEditTypes.RADIO,
+          options: [
```
aaron-downing, line 79: Might be worth adding a validator to this:
`validators: [{ name: ValidatorNames.Required }]`

src/app/forms/templates/psv/psv-dda.template.ts
```diff
@@ -13,73 +14,139 @@ export const PsvDdaTemplate: FormNode = {
           name: 'certificateIssued',
           label: 'DDA certificate issued',
           value: '',
-          type: FormNodeTypes.CONTROL
+          type: FormNodeTypes.CONTROL,
+          viewType: FormNodeViewTypes.STRING,
```
gjulien-bjss, line 18: There is not need to specify `viewType: FormNodeViewTypes.STRING` since it's the default option

src/app/forms/templates/psv/psv-dda.template.ts
```diff
@@ -13,73 +14,139 @@ export const PsvDdaTemplate: FormNode = {
...
+          editType: FormNodeEditTypes.RADIO,
+          options: [
```
aaron-downing, line 20: Same 'Required' validator would be good here too

Ian-Hodges:
Converting to draft. Need to double check backend service interaction. Can't do it right now as working on vt-booking.

src/app/forms/validators/custom-validators.spec.ts
```diff
@@ -258,6 +258,24 @@ describe('numeric', () => {
...
+  it.each([
+    [{ defined: false }, undefined],
+    [null, ''],
+    [null, null],
+    [null, 'hello world!'],
+    [null, 1234],
+  ])('should return %o for %r', (expected: null | {[index: string]: boolean}, input: any) => {
+    const definedValidator = CustomValidators.defined();
+    let form = new FormControl(input)
+    if (typeof input === 'undefined') {
```
andrewf-bjss, line 271: Just checking that `'undefined'` should be the string `'undefined'` (in quotes) here, and that we didn't mean `undefined`

andrewf-bjss (approved):

src/app/forms/validators/custom-validators.ts
```diff
@@ -114,6 +114,15 @@ export class CustomValidators {
     };
   };
 
+  static defined = (): ValidatorFn => {
+    return (control: AbstractControl): ValidationErrors | null => {
+      if (typeof control.value === 'undefined') {
```
gjulien-bjss, line 119: Why not?
```suggestion
      if (control.value === undefined) {
```

gjulien-bjss (approved):

src/app/forms/validators/custom-validators.spec.ts
aaron-downing: Whenever you call typeof on a variable, it returns the type as a string value

src/app/forms/validators/custom-validators.ts
aaron-downing: I read that `typeof` should be used in the case that the variable could potentially not be declared (it's safer), otherwise you would get a reference error by doing a strict `control.value === undefined` check.

PR #540, opened 2022-10-05
fix(cb2-6229): only toggle edit mode when form is valid

aaron-downing:
When a user clicked submit when a form was invalid, the user was switched to the non-edit mode and all changes were lost. 

Now when a form is invalid and the user clicks submit, the user remains on the edit screen and they are scrolled to the top of the page to highlight the errors as per service design

[link to ticket number](https://dvsa.atlassian.net/browse/CB2-6229)

## Checklist

### DIFF
--- a/src/app/features/tech-record/components/edit-tech-record-button/edit-tech-record-button.component.ts
+++ b/src/app/features/tech-record/components/edit-tech-record-button/edit-tech-record-button.component.ts
@@ -6,6 +6,8 @@ import { ofType, Actions } from '@ngrx/effects';
 import { take } from 'rxjs';
 import { Router } from '@angular/router';
 import { GlobalErrorService } from '@core/components/global-error/global-error.service';
+import { GlobalError } from '@core/components/global-error/global-error.interface';
+import { ViewportScroller } from '@angular/common';
 
 @Component({
   selector: 'app-edit-tech-record-button',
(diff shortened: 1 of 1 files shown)


src/app/features/tech-record/components/edit-tech-record-button/edit-tech-record-button.component.ts
```diff
@@ -58,9 +63,16 @@ export class EditTechRecordButtonComponent implements OnInit {
...
+  clickScrollToTop(): void {
+    this.viewportScroller.scrollToPosition([0, 0]);
+  }
+
   submitTechRecord() {
     this.submitCheckFormValidity.emit();
+    this.clickScrollToTop()
 
-    this.toggleEditMode();
+    if (!this.errors) {
```
tom-bjss, line 74: if we're doing this, do we even need to navigate ?

Ian-Hodges (dismissed):

src/app/features/tech-record/components/edit-tech-record-button/edit-tech-record-button.component.ts
tom-bjss: Ignore that, we wont get the updated history without navigating to call the refresh, does this still call the refresh then ?

aaron-downing: you are correct!

tom-bjss (approved):

Ian-Hodges (approved):

PR #544, opened 2022-10-05
chore(cb2-4032): resolve code smells reported by sonarqube

gjulien-bjss:
## Code smells sonarqube

_Resolve some code smells reported by sonarqube_

There still are some issues which will need resolving in either the UI and/or a subsequent PR, however this PR should remove most of the code smells which were found running sonarqube locally
[CB2-4032](https://dvsa.atlassian.net/browse/CB2-4032)

## Checklist


src/app/features/test-records/amend/components/test-amendment-history/test-amendment-history.component.html
```diff
@@ -1,6 +1,8 @@
 <section id="test-records-amendment-history">
-  <h2 class="govuk-heading-m">Test Record Amendment History</h2>
   <table class="govuk-table">
+    <caption class="govuk-table__caption govuk-table__caption--m">
+      Test Record Amendment History
```
mcadenas-bjss, line 4: I think service designs only capitalise first word.

mcadenas-bjss:
```suggestion
      Test record amendment history
```

andrewf-bjss (approved):
Nice work 🙂

sonar-project.properties
```diff
@@ -15,7 +15,18 @@ sonar.sourceEncoding=UTF-8
...
+
+sonar.issue.ignore.multicriteria=e1,e2,e3,e4
+sonar.issue.ignore.multicriteria.e1.ruleKey=typescript:S1128
+sonar.issue.ignore.multicriteria.e1.resourceKey=src/app/api/**
+sonar.issue.ignore.multicriteria.e2.ruleKey=typescript:S3776
+sonar.issue.ignore.multicriteria.e2.resourceKey=src/app/api/**
+sonar.issue.ignore.multicriteria.e3.ruleKey=typescript:S6325
+sonar.issue.ignore.multicriteria.e3.resourceKey=src/app/api/**
+sonar.issue.ignore.multicriteria.e4.ruleKey=typescript:S107
+sonar.issue.ignore.multicriteria.e4.resourceKey=src/app/api/**
```
aaron-downing, line 32: What are these ignoring :eyes:

gjulien-bjss: In order:
1. Allowing unused imports
2. Allowing cognitive complexity above 15
3. Allowing the use of `new RegExp()` instead of the `/regex/` format
4. Removing the limit to max parameters which can be passed to a function

Since these files are generated from the swagger doc, ignoring these specific rules explicitly in these files shouldn't be an issue and improves maintainability. 2 and 4 definitely are not an issue for generated files.

kpearmanbjss (approved):

sonar-project.properties
aaron-downing: No problem, makes sense to ignore. I'm wondering whether we need all the files that have been generated within the api folder as I think a lot of it is boiler plate code we haven't used. I agree the docs are useful but the code I haven't seen us use

Ruairidh-BJSS (approved):

src/app/forms/validators/date/date.validators.ts
```diff
@@ -20,20 +20,26 @@ export class DateValidators {
...
+      return { invalidDate: { error: true, reason: `'${label || 'Date'}' must include time` } };
+    }
+    if (23 < Number.parseInt(hours, 10)) {
+      return { invalidDate: { error: true, reason: `'${label || 'Date'}' hours must be between 0 and 23` } };
+    }
+    if (59 < Number.parseInt(minutes, 10)) {
+      return { invalidDate: { error: true, reason: `'${label || 'Date'}' minutes must be between 0 and 59` } };
+    }
+
+    return null;
```
aaron-downing, line 43: should this method return null? when you have a return null on line 26

gjulien-bjss: I had a "not all code paths return a value" without it. If `includeTime` is true, we can just return `null` here, since the other `if`s before that call would have not been truthy. Without it, I'd probably have to do a check like:
```js
const validTime = this.validateTime(t,label)
if (validTime) {
  return validTime
}
```
And I obviously need the `null` on line 26 is needed in case `validateTime` is false

I do also like that this method is of type `ValidatorFn` which  is `any => ValidationErrors | null` (just like the `validDate` method on line 4), to keep it consistent with the rest of methods in the `custom-validators` folder. Maybe a return type on the function would make it more explicit.

Ruairidh-BJSS (approved):

kpearmanbjss (approved):

PR #546, opened 2022-10-06
fix(cb2-6237): add validation to tech summary dimensions

tahayazici:
## Missing validation of values for dimensions section for a PSV

[CB2-6237](https://dvsa.atlassian.net/browse/CB2-6237)

### DIFF
--- a/src/app/forms/templates/general/dimensions.template.ts
+++ b/src/app/forms/templates/general/dimensions.template.ts
@@ -1,3 +1,4 @@
+import { ValidatorNames } from '@forms/models/validators.enum';
 import { AxleSpacing, VehicleTypes } from '@models/vehicle-tech-record.model';
 import { FormNode, FormNodeEditTypes, FormNodeTypes, FormNodeViewTypes } from '../../services/dynamic-form.types';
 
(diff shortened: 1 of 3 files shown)


naathanbrown (approved):

src/app/forms/templates/general/dimensions.template.ts
```diff
@@ -18,23 +19,26 @@ export function getDimensionsSection(vehicleType: VehicleTypes, noOfAxles: numbe
             value: '',
             type: FormNodeTypes.CONTROL,
             viewType: FormNodeViewTypes.HIDDEN,
-            editType: FormNodeEditTypes.HIDDEN
+            editType: FormNodeEditTypes.HIDDEN,
+            validators: [{ name: ValidatorNames.Max, args: 99999 }]
```
aaron-downing, line 23: I think the `Max` validator creates the issue where it says the value must be less than 99999 which in this case the value can be 99999, `MaxLength` would be the better option. 
image

tahayazici: You are right. Thanks for pointing this out. The error messages have been corrected.

naathanbrown: Ah this was in my PR too

Ruairidh-BJSS (approved):

aaron-downing (approved):

Ian-Hodges (approved):

PR #570, opened 2022-10-13
feat(CB2-5703): Amend HGV & TRL Type Approval section

HazelWilderspin:
## Amend HGV & TRL Type Approval section

Added select to dynamic form, reusable enum map to utils, added validators to type approval and removed gov style override. 
https://dvsa.atlassian.net/browse/CB2-5703

## Checklist

### DIFF
--- a/src/app/forms/components/select/select.component.html
+++ b/src/app/forms/components/select/select.component.html
@@ -1,5 +1,5 @@
 <div *ngIf="isEditing; else view" [ngClass]="{ 'govuk-form-group': !noBottomMargin }">
-  <label *ngIf="label" class="govuk-label" for="{{ name }}"> {{ label }} </label>
+  <label *ngIf="label" class="govuk-label govuk-label--m" for="{{ name }}"> {{ label }} </label>
 
   <div *ngIf="hint" id="{{ name }}-hint" class="govuk-hint">{{ hint }}</div>
 
--- a/src/app/forms/services/dynamic-form.types.ts
+++ b/src/app/forms/services/dynamic-form.types.ts
@@ -69,10 +69,6 @@ export interface FormNodeOption<T> {
   hint?: string;
 }
 
-export interface FormNodeSelectMap {
-  [key: string]: any;
-}
-
 export interface FormNode {
   name: string;
   children?: FormNode[];
@@ -84,7 +80,7 @@ export interface FormNode {
   delimited?: { regex?: string; separator: string };
   value?: any;
   path?: string;
-  options?: FormNodeOption<string | number | boolean | null>[] | FormNodeCombinationOptions | FormNodeSelectMap;
+  options?: FormNodeOption<string | number | boolean | null>[] | FormNodeCombinationOptions;
   validators?: { name: ValidatorNames; args?: any }[];
   customValidatorErrorName?: string;
   asyncValidators?: { name: AsyncValidatorNames; args?: any }[];
--- a/src/app/forms/services/dynamic-form.types.ts
+++ b/src/app/forms/services/dynamic-form.types.ts
@@ -69,6 +69,10 @@ export interface FormNodeOption<T> {
   hint?: string;
 }
 
+export interface FormNodeSelectMap {
+  [key: string]: any;
+}
+
 export interface FormNode {
   name: string;
   children?: FormNode[];
@@ -80,7 +84,7 @@ export interface FormNode {
   delimited?: { regex?: string; separator: string };
   value?: any;
   path?: string;
-  options?: FormNodeOption<string | number | boolean | null>[] | FormNodeCombinationOptions | object[];
+  options?: FormNodeOption<string | number | boolean | null>[] | FormNodeCombinationOptions | FormNodeSelectMap;
   validators?: { name: ValidatorNames; args?: any }[];
   customValidatorErrorName?: string;
   asyncValidators?: { name: AsyncValidatorNames; args?: any }[];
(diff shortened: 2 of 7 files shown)


src/app/forms/templates/general/approval-type.template.ts
```diff
@@ -10,27 +12,47 @@ export function getTypeApprovalSection(vehicleType: VehicleTypes): FormNode {
       {
```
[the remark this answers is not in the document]
HazelWilderspin: Nah, I verified with Nick Choudhury, the boxes needed to match the varchar limit

src/app/forms/services/dynamic-form.types.ts
```diff
@@ -79,7 +80,7 @@ export interface FormNode {
   delimited?: { regex?: string; separator: string };
   value?: any;
   path?: string;
-  options?: FormNodeOption<string | number | boolean | null>[] | FormNodeCombinationOptions;
+  options?: FormNodeOption<string | number | boolean | null>[] | FormNodeCombinationOptions | object[];
```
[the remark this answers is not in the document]
HazelWilderspin: The options prop should be an array of objects taken from the enums, if you want to give me a hand trying to resolve a better way I'd appreciate it

src/app/forms/components/select/select.component.html
```diff
@@ -1,5 +1,5 @@
 <div *ngIf="isEditing; else view" [ngClass]="{ 'govuk-form-group': !noBottomMargin }">
-  <label *ngIf="label" class="govuk-label" for="{{ name }}"> {{ label }} </label>
+  <label *ngIf="label" class="govuk-label govuk-label--m" for="{{ name }}"> {{ label }} </label>
```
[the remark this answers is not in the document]
HazelWilderspin: Ok, I'll look at this one

src/app/forms/services/dynamic-form.types.ts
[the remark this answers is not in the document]
HazelWilderspin: I mimicked the number input, text-are and text-input labels to make the label match everything it was grouped with. The only other use cases of the select component are in the brakes component where the change seems to have effect on it and the defect component. What's the best way to click to a defect to see what it looks like?

src/app/forms/components/select/select.component.html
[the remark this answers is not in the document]
HazelWilderspin: I mimicked the number input, text-are and text-input labels to make the label match everything it was grouped with. The only other use cases of the select component are in the brakes component where the change seems to have effect on it and the defect component. What's the best way to click to a defect to see what it looks like?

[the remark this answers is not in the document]
gjulien-bjss: You could try adding a 6.2 (a) (i) defect for example

src/app/forms/services/dynamic-form.types.ts
```diff
@@ -79,7 +84,7 @@ export interface FormNode {
   delimited?: { regex?: string; separator: string };
   value?: any;
   path?: string;
-  options?: FormNodeOption<string | number | boolean | null>[] | FormNodeCombinationOptions;
+  options?: FormNodeOption<string | number | boolean | null>[] | FormNodeCombinationOptions | FormNodeSelectMap;
```
aaron-downing, line 87: Instead of introducing a new interface, is it not possible to use the existing `FormNodeOption<string | number | boolean | null>[] | FormNodeCombinationOptions` union type on `options` as your return type in your new `getOptionsFromObject` function, as we should want to match what options is expecting instead of extending it's type.

e.g.
`getOptionsFromObject(object: object): FormNodeOption<string | number | boolean | null>[] | FormNodeCombinationOptions`

src/app/forms/utils/enum-map.ts
```diff
@@ -0,0 +1,5 @@
+export default function getOptionsFromObject(object: object): object[] {
+  return Object.values(object).map((value) => {
+      return {value: value, label: value}
+  })
+}
```
gjulien-bjss: - There is a `MultiOptions` interface better suited to the return type of the function
- Can be a bit more concise by wrapping the object in `()` so you can return on the same line
- You can just but `value` in your object because it has the same name as your property

```suggestion
import { MultiOptions } from '@forms/models/options.model';

export default function getOptionsFromObject(object: object): MultiOptions {
  return Object.values(object).map(value => ({value, label: value}))
}
```

src/app/forms/services/dynamic-form.types.ts
naathanbrown: We just did this together

naathanbrown: What do you think of the name? I think it should have something like getSelectOptionsFromObject?

naathanbrown (dismissed):

src/app/forms/services/dynamic-form.types.ts
HazelWilderspin: Nathan helped resolve by terrible interface ^_^;

gjulien-bjss: Also for future reference the interface was not needed as it was the same thing as `Record<string,any>` 😉

src/app/forms/utils/enum-map.ts
HazelWilderspin: eeeerrrm, 
Nathan just helped me reformat that to:

export default function getOptionsFromObject(object: object): FormNodeOption<string | number | boolean>[] {
  return Object.values(object).map(value => {
    return { value: value, label: value };
  });
}

Now IDK what to do

naathanbrown: As far as I can tell the return types we have given you are identical, one is just pre-created for this, I'd maybe move it over to use that type rather than the declared one I gave. In all honestly, I just didn't know this existed

HazelWilderspin: Ok, I used the new one, and it removed your tick.. QQ

gjulien-bjss (approved):

aaron-downing (approved):

PR #574, opened 2022-10-13
chore(*): add more tests for edit tech record button

aaron-downing:
Adds more tests to the edit tech record button, mainly todo with router navigation and user journeys when going to cancel an amendment

### DIFF
--- a/src/app/features/tech-record/components/edit-tech-record-button/edit-tech-record-button.component.spec.ts
+++ b/src/app/features/tech-record/components/edit-tech-record-button/edit-tech-record-button.component.spec.ts
@@ -1,6 +1,6 @@
 import { ComponentFixture, fakeAsync, TestBed, tick } from '@angular/core/testing';
 import { TechRecordModel, VehicleTechRecordModel } from '@models/vehicle-tech-record.model';
-import { provideMockStore } from '@ngrx/store/testing';
+import { MockStore, provideMockStore } from '@ngrx/store/testing';
 import { initialAppState } from '@store/.';
 import { EditTechRecordButtonComponent } from './edit-tech-record-button.component';
 import { By } from '@angular/platform-browser';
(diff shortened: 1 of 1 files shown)


src/app/features/tech-record/components/edit-tech-record-button/edit-tech-record-button.component.spec.ts
```diff
@@ -41,174 +47,189 @@ describe('EditTechRecordButtonComponent', () => {
...
+      fixture.detectChanges();
+  
+      const button = fixture.debugElement.query(By.css('#edit'));
+      
+      if (expected) {
+        expect(button).toBeTruthy;
+      }
+      else {
+        expect(button).toBeFalsy;
+      }
```
gjulien-bjss, line 79: Thought you had to call the functions ?  Does it work like this as well ?
```suggestion
      if (expected) {
        expect(button).toBeTruthy();
      }
      else {
        expect(button).toBeFalsy();
      }
```

Also, detail but: 

```suggestion
      expected ? expect(button).toBeTruthy() : expect(button).toBeFalsy()
```

tahayazici: Agree that these need to be methods (end with brackets), but not with the second suggestion. The ternary conditional operator returns a value. In our case, we just want to call some `void` methods.

gjulien-bjss: There are loads of places where we use it that way though? such as [here](https://github.com/dvsa/cvs-app-vtm/blob/295bd1fcd4db0c9d90acf345694bbc6e30b9030b/src/app/forms/components/checkbox-group/checkbox-group.component.ts#L28) and [here](https://github.com/dvsa/cvs-app-vtm/blob/295bd1fcd4db0c9d90acf345694bbc6e30b9030b/src/app/features/tech-record/components/vehicle-technical-record/vehicle-technical-record.component.ts#L77). It's not because it returns something we have to do something with that return value. With this suggestion it will just execute the functions. There are examples online using ternary operators this way such as [this one](https://www.w3schools.com/react/react_es6_ternary.asp) and [this one](https://stackabuse.com/guide-to-the-ternary-operator-in-javascript/)

tahayazici: Ok, nice. If W3Schools references it, that's good enough for me 🙂 Thanks for the links!

But still, this shouldn't mandate we change all our `if/else`s into ternaries. So, I'd leave this up to OP.

tahayazici (approved):

gjulien-bjss (approved):

PR #578, opened 2022-10-14
feat(cb2-6092): hide emissions section

CormacBJSS:
## Ticket title

Hide Emissions fields / certificate fields based on test result
[CB2-6092](https://dvsa.atlassian.net/browse/CB2-6092)

## Checklist

### DIFF
--- a/src/app/features/test-records/components/base-test-record/base-test-record.component.ts
+++ b/src/app/features/test-records/components/base-test-record/base-test-record.component.ts
@@ -68,7 +68,7 @@ export class BaseTestRecordComponent implements AfterViewInit {
     return this.testRecordsService.sectionTemplates$;
   }
 
-  get editingTestResult$(): Observable<string | undefined> {
+  get resultOfTest$(): Observable<string | undefined> {
     return this.testRecordsService.editingTestResult$.pipe(map(testResult => testResult?.testTypes[0].testResult));
   }
 
--- a/src/app/features/test-records/components/base-test-record/base-test-record.component.ts
+++ b/src/app/features/test-records/components/base-test-record/base-test-record.component.ts
@@ -68,6 +68,10 @@ export class BaseTestRecordComponent implements AfterViewInit {
     return this.testRecordsService.sectionTemplates$;
   }
 
+  get editingTestResult$(): Observable<string | undefined> {
+    return this.testRecordsService.editingTestResult$.pipe(map(testResult => testResult?.testTypes[0].testResult));
+  }
+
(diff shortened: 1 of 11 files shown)


src/app/features/test-records/components/base-test-record/base-test-record.component.ts
```diff
@@ -68,6 +68,10 @@ export class BaseTestRecordComponent implements AfterViewInit {
     return this.testRecordsService.sectionTemplates$;
   }
 
+  get editingTestResult$(): Observable<string | undefined> {
```
gjulien-bjss, line 71: Should probably be called something like `resultOfTest$` to match the rest of the application. Especially since `editingTestResult$` is something else in the component

src/app/forms/validators/custom-async-validators.spec.ts
```diff
@@ -114,5 +115,103 @@ describe('updateTestStationDetails', () => {
...
+
+    form = new FormGroup({
+      foo: new CustomFormControl({ name: 'foo', type: FormNodeTypes.CONTROL, children: [] }, null)
+    });
+  });
+
+  it('should be required when value is "pass"', async () => {
+    form.controls['foo'].patchValue('');
+
+    const testResult = mockTestResult();
```
gjulien-bjss, line 141: if all you need some the `testResult` here is the result of the test i.e.: pass, fail etc..., you can do 
```js
const testResult = {testTypes: [{testResult: resultOfTestEnum.pass}]} as TestResultModel 
```
instead of passing the whole `mockTestResult()` everytime. And if you need only 2-3 more properties, you should be able to pass them in the same fashion.

If you need much more than 2-3 properties in the test, you could consider assigning a variable in the `beforeEach` just as 
```js
let mockTestResult = mockTestResult()
```
and then change each testResult to the resultOfTest you need in each test

gjulien-bjss (commented):
A couple minor comments, a lot of linting errors, consider turning "format on save" on in VSCode, should help

src/app/forms/validators/custom-async-validators.spec.ts
CormacBJSS: Thanks that change has been made.

src/app/features/test-records/components/base-test-record/base-test-record.component.ts
CormacBJSS: Thanks that change has been made.

kpearmanbjss (approved):

gjulien-bjss (approved):

PR #595, opened 2022-10-20
feat(CB2-5695): Amend PSV Vehicle Summary section for provisional record

HazelWilderspin:
Amend PSV Vehicle Summary section for provisional record

Added: styles, validators, enum, please see extended notes in comments
[link to ticket number](https://dvsa.atlassian.net/browse/CB2-5695)

Converted to draft until parking brake fiasco is resolved

## Checklist


HazelWilderspin (commented):
Extended notes:

- Ignored Axles: needs adding to CB2-6095: Amend number of axles. This ticket will determine how data is handled and the UI needs to fit around it.
- Also, the Boolean values don't marry up to the "Front 2 axles" display. If this is required, then that also needs handling with data management.
- Vehicle class code referenced in data requirements does not exist on the json object, also not mentioned in the UI reference
- Euro standard appears unchecked in edit mode because the incoming data is incorrect, "m1", which is not a matching string to any of the enum values. Permanent change of mock data recommended.
- Seats: No UI reference for non-edit mode, changing the labels + adding a title would affect both
- Float: New ID prop and style is reusable for any node in dynamic form, it is now possible to cherry pick a component to make changes
- Approval type: demanded by API but data value is "". When doing approval type ticket, it specified HGV and TRL, not sure how to resolve.

src/app/forms/templates/general/approval-type.template.ts
```diff
@@ -15,9 +15,9 @@ export function getTypeApprovalSection(vehicleType: VehicleTypes): FormNode {
         type: FormNodeTypes.CONTROL,
         editType: FormNodeEditTypes.SELECT,
         options: getOptionsFromEnum(approvalType),
-        validators: [
+        validators: (vehicleType === VehicleTypes.PSV ? [] : [
```
naathanbrown, line 18: Same as other PR, think this is mandatory for all vehicles

src/app/models/fuel-propulsion-system.model.ts
```diff
@@ -0,0 +1,9 @@
+export enum FuelPropulsionSystem {
```
naathanbrown, line 1: This exists as 'FuelTypes' on line 26 of vehicle-tech-record.model.ts

src/app/forms/templates/psv/psv-tech-record.template.ts
```diff
@@ -119,91 +142,115 @@ export const PsvTechRecord: FormNode = {
...
+      viewType: FormNodeViewTypes.STRING,
+      editType: FormNodeEditTypes.DATE,
+      isoDate: false
+    },
+    {
+      name: 'vehicleSize',
+      label: 'Vehicle size',
+      value: '',
+      type: FormNodeTypes.CONTROL,
+      viewType: FormNodeViewTypes.STRING,
```
naathanbrown, line 239: Don't need to declare this on any that is a string, it defaults to this I believe

naathanbrown:
> Extended notes:
> 
> > * Ignored Axles: needs adding to CB2-6095: Amend number of axles. This ticket will determine how data is handled and the UI needs to fit around it.
> > * Also, the Boolean values don't marry up to the "Front 2 axles" display. If this is required, then that also needs handling with data management.
> > * Vehicle class code referenced in data requirements does not exist on the json object, also not mentioned in the UI reference
> > * Euro standard appears unchecked in edit mode because the incoming data is incorrect, "m1", which is not a matching string to any of the enum values. Permanent change of mock data recommended.
> > * Seats: No UI reference for non-edit mode, changing the labels + adding a title would affect both
> > * Float: New ID prop and style is reusable for any node in dynamic form, it is now possible to cherry pick a component to make changes
> 
> Done:
> 
> > * Approval type: demanded by API but data value is "". When doing approval type ticket, it specified HGV and TRL, not sure how to resolve.

> Axle work fine, also the parking break stuff sould be done in that ticket, leave as is for now.
> It is in the JSON, but have confirmed that it adjusts itself when you change the box

src/app/forms/templates/general/approval-type.template.ts
HazelWilderspin: Ok cool, that gets rid of one extended note...
Are they the same values?
Or does it need a different enum?

src/app/forms/templates/psv/psv-tech-record.template.ts
```diff
@@ -9,33 +16,34 @@ export const PsvTechRecord: FormNode = {
...
+      editType: FormNodeEditTypes.DATE,
+      validators: [{ name: ValidatorNames.Required }],
+      isoDate: false
     },
     {
       name: 'manufactureYear',
       label: 'Year of manufacture',
       value: '',
-
+      width: FormNodeWidth.XS,
```
aaron-downing, line 36: I'm not sure whether this width is actually working 
image

src/app/forms/templates/psv/psv-tech-record.template.ts
```diff
@@ -46,7 +54,6 @@ export const PsvTechRecord: FormNode = {
           name: 'dtpNumber',
           label: 'DTP number',
           value: '',
```
aaron-downing, line 56: Could do with width adding for this box

image

src/app/forms/templates/psv/psv-tech-record.template.ts
```diff
@@ -119,91 +137,105 @@ export const PsvTechRecord: FormNode = {
...
+      editType: FormNodeEditTypes.DATE,
+      isoDate: false
+    },
+    {
+      name: 'vehicleSize',
+      label: 'Vehicle size',
+      value: '',
+      type: FormNodeTypes.CONTROL,
+      editType: FormNodeEditTypes.RADIO,
+      options: getOptionsFromEnum(VehicleSize)
```
aaron-downing, line 227: Very minor but could we capitalise `small` and `large` please
image

src/app/forms/templates/psv/psv-tech-record.template.ts
```diff
@@ -9,33 +16,34 @@ export const PsvTechRecord: FormNode = {
...
-
+      width: FormNodeWidth.XS,
       type: FormNodeTypes.CONTROL,
-      viewType: FormNodeViewTypes.STRING
+      editType: FormNodeEditTypes.NUMBER,
+      validators: [{ name: ValidatorNames.MaxLength, args: 4 }, { name: ValidatorNames.Required }]
     },
     {
       name: 'noOfAxles',
       label: 'Number of axles',
```
aaron-downing, line 43: Also wondering whether we should be displaying noOfAxles within vehicle summary along with axle fitted with parking brake? Might be one to ask service design where it belongs
image

aaron-downing: After reading your comment above, I think it's worth us removing the axle fields in case we forget to down the line, plus it means that we're matching up with intended service design and saying that vehicle summary section is done without having to revisit it in future

aaron-downing:
For the seat related fields could we add a `Seats` subheading like the service design show? Also it seems like it's trying to render something between the `Seats X deck` and `Standing capacity` I'm not sure whether that's to do with the new custom component or something 
image

src/app/forms/templates/psv/psv-tech-record.template.ts
HazelWilderspin: fixt

src/app/forms/templates/psv/psv-tech-record.template.ts
HazelWilderspin: I left them out because they were out of scope and last I heard they wanted a dropdown for number of axles, IDK if that's true anymore

src/app/forms/templates/psv/psv-tech-record.template.ts
aaron-downing: (realise that axle fitted with parking brake is on service design, not sure what approach to take for what they have on that service design, I think it's worth a conversation)

src/app/forms/templates/psv/psv-tech-record.template.ts
HazelWilderspin: Err I don't think so no, because the options come from an enum and the auto select is case sensitive

src/app/forms/templates/psv/psv-tech-record.template.ts
aaron-downing: Sorry I meant for dtpNumber

src/app/forms/templates/psv/psv-tech-record.template.ts
```diff
@@ -61,15 +68,14 @@ export const PsvTechRecord: FormNode = {
       children: [
         {
           name: '0',
-          label: 'Axel',
           value: '',
           type: FormNodeTypes.GROUP,
           children: [
             {
               name: 'parkingBrakeMrk',
-              label: 'Axles fitted with a parking brake',
```
aaron-downing, line 70: I think we might need some guidance with this one as the ticket says it's an optional boolean but the backend is saying it's a required field, so we might need to update the backend in a similar way we did for dda ceritifcateIssued. I think it'd be good to speak with service design though about what the options are in this dropdown as it seems confusing from the ticket how it would work
image

src/app/forms/templates/psv/psv-tech-record.template.ts
gjulien-bjss: You could use the `DefaultNullOrEmpty` pipe in the component on your options, which should capitalise it without affecting the underlying value. Or  you could pass in some options such as 
```js
   options: [
        { value: 'small', label: 'Small' },
        { value: 'large', label: 'Large' }
      ]
```
and not use the enum to achieve this

gjulien-bjss: Or change the `getOptionsFromEnum` function to capitalise the first letter of the `label` field

HazelWilderspin: "Or change the getOptionsFromEnum function to capitalise the first letter of the label field"

This is smart, I did this

src/app/forms/components/dynamic-form-group/dynamic-form-group.component.html
```diff
@@ -55,7 +59,7 @@ <h1 class="govuk-heading-xl title">{{ form.meta.label }}</h1>
         [formControlName]="control.key"
         class="govuk-summary-list"
       ></div>
-      <app-dynamic-form-field *ngSwitchCase="true" [control]="control" [form]="formGroup"></app-dynamic-form-field>
+      <app-dynamic-form-field *ngSwitchCase="true" [control]="control" [form]="formGroup" [attr.id]="control.value.meta.id"></app-dynamic-form-field>
```
mcadenas-bjss, line 62: Have you considered all the use cases for this `id`? When generating form arrays, there would be duplicate IDs in the DOM which should be avoided. Is having an id on the `app-dynamic-form-field` tag meaningful? The id would probably be more useful on the actual input/element with the value. Wick all that said. I think there are still instances of duplicate IDs due to the way dynamic forms are generated. Might be worth fixing that instead.

mcadenas-bjss (approved):

gjulien-bjss (approved):

PR #606, opened 2022-10-24
feat(cb2-6062) amend a current technical record

tom-bjss:
purpose of this pr is to be able to amend a current record for erranous values or continue with the old journey for notifyable alterations.

This required a new page adding to the journey with adio buttons similar to test reuslts.

The new journeys are as follows 

Tech Record (view) -> Click ammend button -> New radio button page -> Amend Current -> Navigation to Tech record with edit mode & option 1

Tech Record (view) -> Click ammend button -> New radio button page -> Notifyable alteration-> Navigation to Tech record with edit mode & option 2

NOTE: theres a todo comment that I would like you to comment on

I have refactored checking the status code in the technical record service and this was implemented in a way to only work for provisional records


src/app/features/tech-record/components/edit-tech-record-button/edit-tech-record-button.component.spec.ts
```diff
@@ -183,10 +193,10 @@ describe('EditTechRecordButtonComponent', () => {
             fixture.detectChanges();
             fixture.debugElement.query(By.css('#cancel')).nativeElement.click();
 
-            expect(router.navigate).toHaveBeenCalledWith([]);
```
tom-bjss, line 186: Had to change this one sadly, as it was becoming a pain point to refactor while using active route

tahayazici (approved):

HazelWilderspin:
If I amend a record and then try to amend again I get 404 not found.. and if I empty a field that I expect to be required by the API then the submit button doesn't trigger any UI errors?

HazelWilderspin:
Also, the intermediary page only appears for the trl vin I have, is it supposed to appear for hgvs and psvs?

naathanbrown:
I cannot make this show me anything other than a 404 when I click amend and then pick either option... This was tested with VIN: 	XXB6703742N122212

naathanbrown:
This is the path I get - http://localhost:4200/tech-records/3506666/XXB6703742N122212/amend-reason/notifiable-alteration-needed

HazelWilderspin:
Is that because I got in there first and made an amendment on all 3 vins, Effectively locking you out of all edit capability?

tom-bjss:
> If I amend a record and then try to amend again I get 404 not found.. and if I empty a field that I expect to be required by the API then the submit button doesn't trigger any UI errors?

This is because youre on the historic url now which is a separate issue all together
That second one ill look into :D

aaron-downing (approved):

tahayazici (approved):

tom-bjss:
Live demo to approvers

PR #614, opened 2022-10-25
fix(cb2-6062): added default selection to radio group

tom-bjss:
added default value to radio group to reflect new enum as this was based on magic numbers before, this will stop users from being able to press continue without a valid reason selected

### DIFF
--- a/src/app/features/tech-record/components/tech-amend-reason/tech-amend-reason.component.ts
+++ b/src/app/features/tech-record/components/tech-amend-reason/tech-amend-reason.component.ts
@@ -4,6 +4,7 @@ import { Router, ActivatedRoute } from '@angular/router';
 import { GlobalErrorService } from '@core/components/global-error/global-error.service';
 import { CustomFormControl, CustomFormGroup, FormNodeOption, FormNodeTypes } from '@forms/services/dynamic-form.types';
 import { ReasonForEditing } from '@models/vehicle-tech-record.model';
+import { GlobalError } from '@core/components/global-error/global-error.interface';
 
 
 @Component({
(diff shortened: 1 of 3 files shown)


src/app/features/tech-record/components/tech-amend-reason/tech-amend-reason.component.ts
```diff
@@ -23,14 +23,13 @@ export class TechAmendReasonComponent {
 
     this.form = new CustomFormGroup(
       { name: 'reasonForAmend', type: FormNodeTypes.GROUP },
-      { reason: new CustomFormControl({ name: 'reason', type: FormNodeTypes.CONTROL }, 2, [Validators.required]) }
+      { reason: new CustomFormControl({ name: 'reason', type: FormNodeTypes.CONTROL }, 'correcting-an-error', [Validators.required]) }
```
mcadenas-bjss, line 26: If this is to match new enum, would be better to use the enum here when initialising the control value.

Ian-Hodges (commented):
Are we okay preselecting an option in a radio group? It's generally something to avoid.
https://design-system.service.gov.uk/components/radios/

mcadenas-bjss (approved):

Ian-Hodges (dismissed):

kpearmanbjss (approved):

aaron-downing (approved):

aaron-downing:
🚀

aaron-downing:
Cosmetic change that's probably not for this PR but would be nice to have some breadcrumbs to be able to navigate to the tech record you were viewing instead of just the home screen or having to click the back button.
image

mcadenas-bjss (approved):

PR #616, opened 2022-10-26
fix(*): clear errors on navigation and stop errors in tech record non-edit mode

aaron-downing:
Changes:
- Removes persistence of errors across pages, errors will now be cleared whenever user navigates through VTM 
- Errors will no longer be displayed on a tech record in non-edit mode that were previously caused by values on the tech record not matching the validations we have on the form.

### DIFF
--- a/src/app/features/tech-record/components/vehicle-technical-record/vehicle-technical-record.component.spec.ts
+++ b/src/app/features/tech-record/components/vehicle-technical-record/vehicle-technical-record.component.spec.ts
@@ -130,6 +130,7 @@ describe('VehicleTechnicalRecordComponent', () => {
       component.vehicleTechRecord = mockVehicleTechnicalRecord();
       fixture.detectChanges();
 
+      component.isEditing = true;
       component.handleSubmit()
 
       expect(handleFormStateSpy).toHaveBeenCalled();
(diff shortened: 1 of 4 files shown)


naathanbrown (approved):

tom-bjss (dismissed):

tahayazici (approved):

src/app/core/components/global-error/global-error.service.ts
```diff
@@ -12,8 +13,13 @@ import { GlobalError } from './global-error.interface';
 export class GlobalErrorService {
   private errors: Observable<GlobalError[]>;
 
-  constructor(private store: Store<State>) {
+  constructor(private store: Store<State>, private router: Router) {
     this.errors = this.store.pipe(select(globalErrorState));
+    this.router.events.subscribe((event)=> {
```
gjulien-bjss, line 18: Not super necessary, and I don't know if it's a better implementation, but the ngrx route store package dispatches actions when the user navigates. Maybe creating an effect which listens to these actions and clears the errors could be a cleaner/better implementation. Just a suggestion mostly
ngrx

gjulien-bjss (approved):

src/app/core/components/global-error/global-error.service.ts
mcadenas-bjss: Global errors reducer might be able to act on the `navigated` action and clear errors.

mcadenas-bjss (approved):
Looks good but I like @gjulien-bjss suggestion.

src/app/core/components/global-error/global-error.service.ts
aaron-downing: I'll look into this, it's a good suggestion

gjulien-bjss (approved):

tom-bjss (approved):

PR #618, opened 2022-10-26
feat(cb2-6092): customise emissions sections

CormacBJSS:
## Ticket title

Customise Emissions section / certificate fields based on test result
[https://dvsa.atlassian.net/browse/CB2-6092]

## Checklist

### DIFF
--- a/src/app/forms/validators/custom-async-validators.ts
+++ b/src/app/forms/validators/custom-async-validators.ts
@@ -70,14 +70,14 @@ export class CustomAsyncValidators {
     };
   }
 
-  static requiredIfNotResult(store: Store<State>, result: any): AsyncValidatorFn {
+  static requiredIfNotResult(store: Store<State>, result: resultOfTestEnum | resultOfTestEnum[]): AsyncValidatorFn {
     return (control: AbstractControl): Observable<ValidationErrors | null> =>
       store.pipe(
         take(1),
         select(testResultInEdit),
         map(testResult => {
           const currentResult = testResult?.testTypes[0].testResult;
-          if ((Array.isArray(result) ? !result.includes(currentResult) : currentResult !== result) && (control.value == null || control.value === '')) {
+          if ((Array.isArray(result) ? currentResult&&!result.includes(currentResult) : currentResult !== result) && (control.value == null || control.value === '')) {
             if(Array.isArray(result)) return { requiredIfNotResult: true }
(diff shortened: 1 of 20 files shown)


leifkemp-bjss (commented):
Could probably do with unit tests for the validator and error messages

kpearmanbjss (dismissed):

src/app/forms/validators/custom-async-validators.ts
```diff
@@ -70,15 +70,16 @@ export class CustomAsyncValidators {
     };
   }
 
-  static requiredIfNotResult(store: Store<State>, result: resultOfTestEnum): AsyncValidatorFn {
+  static requiredIfNotResult(store: Store<State>, result: any): AsyncValidatorFn {
```
gjulien-bjss, line 73:
```suggestion
  static requiredIfNotResult(store: Store<State>, result: resultOfTestEum | resultOfTestEnum[]): AsyncValidatorFn {
```

src/app/forms/validators/custom-async-validators.ts
```diff
@@ -92,4 +93,28 @@ export class CustomAsyncValidators {
...
+            if ((Array.isArray(result) ? !result.includes(currentResult) : currentResult !== result) && newValue && (control.value === null || control.value === undefined || control.value === '')) {
+              return { requiredIfNotResultAndSiblingEquals: { sibling: siblingControl.meta.label } };
+            }
+          }
+
+          return null;
+        })
+      );
+  }
+
```
gjulien-bjss, line 119: Why can you not attach the `requiredIfSiblingEquals` validator (or create it if it doesn't exist) and the `requiredIfResult` validators on the control?

CormacBJSS: What I am trying to achieve is rules like: "particulateTrapFitted should only be required if modType code is p and the result is pass"

The problem is the validators are independent of each other so there is no way to do an "AND" condition as far as I know. If either validator fails then the error message is shown. Is there a way around that?

src/app/forms/validators/custom-async-validators.ts
CormacBJSS: That looks good thanks.

kpearmanbjss (approved):

mcadenas-bjss (approved):

PR #620, opened 2022-10-27
feature(cb2-5704): add the body section for hgv and trl

tahayazici:
## Amend HGV & TRL Body section

- Create a custom body component to replace the default existing one.
- Make the `select` component resizable.
- Apply minor corrections to the `ngOnChanges` of various custom sections.
- Populate extra mock reference data for body make and model.
- Create enums for `BodyTypeDescription` and `BodyTypeCode` alongside helper methods that aid in retrieving those in the right format.

[CB2-5704](https://dvsa.atlassian.net/browse/CB2-5704)


src/app/forms/components/select/select.component.scss
```diff
@@ -0,0 +1,32 @@
...
+
+.govuk-input--width-20 {
+  max-width: 20.5em !important;
+}
+
+.govuk-input--width-10 {
+  max-width: 11.5em !important;
+}
+
+.govuk-input--width-5 {
```
aaron-downing, line 18: Are these overriding the existing govuk-input--width classes?

tahayazici: The `govuk-select` class is setting the `max-width` which is not allowing us to resize it. Here I am appending the `!important` tag to the `govuk-input-width` classes so that the resizing works.

aaron-downing (commented):
Can we get agreement from BA/POs that we're going to be using mock data and update the ticket to reflect that as we will likely need an additional ticket in the backlog to hook this up to reference data.

src/app/forms/templates/general/body.template.ts
```diff
@@ -1,39 +1,35 @@
-import { MultiOptionsService, SpecialRefData } from '@forms/services/multi-options.service';
-import getOptionsFromEnum from '@forms/utils/enum-map';
+import { ValidatorNames } from '@forms/models/validators.enum';
```
naathanbrown, line 1: We spoke a while ago about slowly removing any code from templates and making them pure JSON, this isn't mandatory but does make sense to do them as we go if possible

naathanbrown:
Unsure about the function code and body type code here? In non edit mode function is displaying 'A' which appears to be a body type code. Is this for sure working correctly and the tech record is being updated correctly?

src/app/models/body-type-enum.ts
```diff
@@ -0,0 +1,69 @@
...
+  M = 'm',
+  O = 'o',
+  P = 'p',
+  R = 'r',
+  S = 's',
+  T = 't',
+  X = 'x'
+}
+
+export function getCode(bodyType: BodyTypeDescription): BodyTypeCode {
```
aaron-downing, line 42: Could you create a hash map instead of this long switch case statement

tom-bjss: Seconded

src/app/models/body-type-enum.ts
```diff
@@ -0,0 +1,69 @@
...
+  PETROL_OR_OIL_TANKER = 'petrol/oil tanker',
+  REFUSE = 'refuse',
+  REFRIGERATED = 'refrigerated',
+  SINGLE_DECKER = 'single decker',
+  SKELETAL = 'skeletal',
+  SKIP_LOADER = 'skip loader',
+  TIPPER = 'tipper',
+}
+
+export function getBodyTypesAsOptions(): MultiOptions {
```
aaron-downing, line 20: We have a function for this here https://github.com/dvsa/cvs-app-vtm/blob/develop/src/app/forms/utils/enum-map.ts

tahayazici: Many thanks, using this instead now.

src/app/models/body-type-enum.ts
tahayazici: I knew there was a cleaner way to achieve this but couldn't think of any at the time. Thanks for the great suggestion!

src/app/forms/templates/general/body.template.ts
tahayazici: There are still at a lot of templates that encapsulate some switching logic, but ok, sure.

`body.template.ts` is now deleted and instead broken down into:
- `psv-body.template.ts`
- `hgv-trl-body.template.ts`

tahayazici:
To address your curiosity @naathanbrown, body type and code are within the `bodyType` object as such:
`bodyType: { description: string, code: string }`

If you check the JIRA ticket, the `functionCode` is expected to be a single-digit number field. I also noticed that before submitting any changes, it comes back as "A" initially (VIN: H00037502), but that could be an error in the data, I think it's best to stick to the ticket's requirements.

tahayazici:
@aaron-downing we had a discussion in the mod call about this. All previous tickets needing reference data mocked it. There is currently a ticket that the test team are working on to enable seeding reference data. I will link that to my ticket.

src/app/forms/templates/general/hgv-trl-body.template.ts
```diff
@@ -0,0 +1,52 @@
+import { ValidatorNames } from '@forms/models/validators.enum';
+import { FormNode, FormNodeEditTypes, FormNodeTypes } from '../../services/dynamic-form.types';
+
+export const hgvAndTrlBodyTemplate: FormNode = {
+  name: 'bodySection',
+  label: 'Body',
+  type: FormNodeTypes.GROUP,
+  children: [
+    {
+      name: 'make',
```
aaron-downing, line 10: Quite a lot of these fields are mandatory, need to add validators.

src/app/forms/templates/general/hgv-trl-body.template.ts
```diff
@@ -0,0 +1,52 @@
...
+      value: '',
+      type: FormNodeTypes.CONTROL,
+      editType: FormNodeEditTypes.SELECT,
+      options: []
+    },
+    {
+      name: 'model',
+      label: 'Body model',
+      value: '',
+      type: FormNodeTypes.CONTROL,
```
aaron-downing, line 21: Missing maxLength validator

aaron-downing:
> @aaron-downing we had a discussion in the mod call about this. All previous tickets needing reference data mocked it. There is currently a ticket that the test team are working on to enable seeding reference data. I will link that to my ticket.

No problem, but I think we need an additional ticket for changing the VTM code to point to the body reference data in order to track it and make sure we don't forget.

src/app/forms/components/select/select.component.scss
aaron-downing: That's so bizarre, I remember us getting it to override `govuk-input` in the `text-input` component

tahayazici:
> > @aaron-downing we had a discussion in the mod call about this. All previous tickets needing reference data mocked it. There is currently a ticket that the test team are working on to enable seeding reference data. I will link that to my ticket.
> 
> No problem, but I think we need an additional ticket for changing the VTM code to point to the body reference data in order to track it and make sure we don't forget.

Sure, I created [CB2-6511](https://dvsa.atlassian.net/browse/CB2-6511) so we can track this on the board.

src/app/forms/templates/general/hgv-trl-body.template.ts
tahayazici: Done

src/app/forms/templates/general/hgv-trl-body.template.ts
tahayazici: Done

naathanbrown (approved):

src/app/forms/templates/general/hgv-trl-body.template.ts
```diff
@@ -0,0 +1,53 @@
...
+          validators: [{ name: ValidatorNames.Required }]
+        },
+      ]
+    },
+    {
+      name: 'functionCode',
+      label: 'Function code',
+      value: '',
+      type: FormNodeTypes.CONTROL,
+      validators: [{ name: ValidatorNames.MaxLength, args: 1 }]
```
aaron-downing, line 43: Just looked at the service design and function code seems to be a dropdown, might be worth seeing if it's intentional or not

src/app/forms/templates/general/hgv-trl-body.template.ts
```diff
@@ -0,0 +1,53 @@
...
+    },
+    {
+      name: 'bodyType',
+      label: 'Body type',
+      value: '',
+      type: FormNodeTypes.GROUP,
+      children: [
+        {
+          name: 'description',
+          label: 'Body type',
```
aaron-downing, line 31: Are we not having a dropdown of just standard values for body type instead of free text? I think the backend expects certain input for this field (screenshot from the BE API code).
image

aaron-downing: Oh wait we're using the multi options service now so the form options aren't shown in the templates right?

tahayazici: All of the values shown on your image match our `BodyTypeDescription` enum. So, it should be fine.

src/app/forms/templates/general/hgv-trl-body.template.ts
tahayazici: I have used the service design and made it a dropdown already 🙂

naathanbrown (approved):

aaron-downing (approved):

PR #628, opened 2022-10-31
fix(cb2-5702): rewrite the dimensions templates

tahayazici:
## Amend HGV & TRL Dimensions section

This pull request is the result of a discovery made with the help of @aaron-downing as not all parts of the dimensions section functioned as expected.

Previously, labelling the axles spacing fields dynamically was very troublesome. So, we decided to manually populate them through explicit code in the dimensions template. This was fine at the time as we were only displaying technical record fields and not editing them. Now that editing is enabled, these form fields do not bind to the object's properties.

Thus...
- The dimensions template is simplified and rewritten from scratch excluding any logic (we manage all in the custom section now).
- The double binding bug between the first and second min-max field groups is fixed.
- Validators have been corrected.

[CB2-5702](https://dvsa.atlassian.net/browse/CB2-5702)


src/app/forms/custom-sections/dimensions/dimensions.component.html
```diff
@@ -51,13 +51,13 @@
     suffix="mm"
   ></app-switchable-input>
 
-  <ng-container *ngIf="!isPsv">
+  <ng-container *ngIf="!isPsv && hasAxleSpacings">
```
Ian-Hodges, line 54: If you have a hgv and it was missing its axelSpacings for some reason, would this prevent the user from adding any? i.e. when editing, the control would not be rendered for them to enter data.

tom-bjss: It would, I cant imagine a world where you wouldn't. The only other piece of data I could imagine using is Axels then

src/app/forms/custom-sections/dimensions/dimensions.component.ts
```diff
@@ -66,11 +71,19 @@ export class DimensionsComponent implements OnInit, OnChanges, OnDestroy {
...
+    return !!this.vehicleTechRecord.dimensions?.axleSpacing?.length;
+  }
+
+  get axleSpacings(): CustomFormGroup {
+    return this.form.get(['dimensions', 'axleSpacing']) as CustomFormGroup;
   }
 
-  get bottomSection(): CustomFormGroup {
-    return this.form.get(['dimensionsBottomSection', 'dimensions']) as CustomFormGroup;
+  getAxleSpacing(i: number): CustomFormGroup {
```
tom-bjss, line 86: I know you're passing in I as an index from a different place in the code, but could this I have a better name ?
index would do

tom-bjss (approved):

src/app/forms/custom-sections/dimensions/dimensions.component.html
Ian-Hodges: Is it expected that all trailers and HGVs have axelSpacings populated? If so you could leave it just as `*ngIf="!isPsv"`?

tahayazici: Initially, the check was only `*ngIf="!isPsv"`. But this breaks some of the unit tests using mock data and some of our data in the DB is also crap, when using a VIN for an incomplete record from the DB, with the `hasAxleSpacings` check, this breaks the page.

So, I will leave this here for completeness' sake and we can remove it in the future when we start working with real data.

aaron-downing:
Should a TRL have this field when TRLs usually only have 1 axle? I know it says it on the ticket but it seems confusing to me
image

tahayazici:
I went with what the ticket suggests @aaron-downing, this should be up to the POs in when this gets in PO Review.

src/app/forms/custom-sections/dimensions/dimensions.component.ts
tahayazici: This same pattern is used in other places too. `index` is too verbose man 🙂

src/app/forms/custom-sections/dimensions/dimensions.component.html
Ian-Hodges: We shouldn't be adding code to work around our dodgy test/mock data. I'm not familiar enough with the data, but if it is possible for a trailer or hgv to be added without any axelSpacings, then it's going to cause problems.
Will the same form be used to create new tech records?

Ian-Hodges (approved):

PR #644, opened 2022-11-04
feat(cb2-6092): hide emissions section

CormacBJSS:
## Hide emissions section and certificate number - new validator

New custom validator to allow the certificate number for Specialist testTypeIds 150, 151, 181, 182 to be hidden - these testTypeIds are part of Specialist Test Group 1 but the certificate field needs to be displayed for the rest of the testTypeIds in Group 1.

[https://dvsa.atlassian.net/browse/CB2-6092]

## Checklist

### DIFF
--- a/src/app/forms/validators/custom-async-validators.ts
+++ b/src/app/forms/validators/custom-async-validators.ts
@@ -139,17 +139,17 @@ export class CustomAsyncValidators {
   }
 
   static checkConditions(testResult: TestResultModel, conditions: Condition | Condition[] ){
-    if(Array.isArray(conditions)){
-      let failed: boolean = false;
-
-      conditions.forEach((condition) => {
-        if(!CustomAsyncValidators.checkCondition(testResult, condition)) failed = true;
-      });
-
-      return !failed;
-    }else{
+    if (!Array.isArray(conditions)) {
       return CustomAsyncValidators.checkCondition(testResult, conditions);
     }
+
+    for (const condition of conditions) {
+      if(!CustomAsyncValidators.checkCondition(testResult, condition)) {
+        return false;
+      }
+    }
+    
+    return true;
   }
 
   static checkCondition(testResult: TestResultModel, condition: Condition ){
--- a/src/app/forms/validators/custom-async-validators.ts
+++ b/src/app/forms/validators/custom-async-validators.ts
@@ -140,7 +140,7 @@ export class CustomAsyncValidators {
       );
   }
 
-  static checkConditions(testResult: TestResultModel, conditions: Condition | Condition[] ){
+  private static checkConditions(testResult: TestResultModel, conditions: Condition | Condition[] ){
     if (!Array.isArray(conditions)) {
       return CustomAsyncValidators.checkCondition(testResult, conditions);
     }
@@ -148,15 +148,15 @@ export class CustomAsyncValidators {
     return conditions.every(condition => CustomAsyncValidators.checkCondition(testResult,condition))
   }
 
-  static checkCondition(testResult: TestResultModel, condition: Condition ){
+  private static checkCondition(testResult: TestResultModel, condition: Condition ){
     const {field, operator, value} = condition
 
     const fieldValue = testResult.testTypes[0].hasOwnProperty(field)
-      ? (testResult?.testTypes[0] as any)[field]
+      ? (testResult.testTypes[0] as any)[field]
       : (testResult as any)[field];
 
     const isTrue = Array.isArray(value) ? value.includes(fieldValue) : fieldValue === value;
 
     return operator === operatorEnum.Equals ? isTrue : !isTrue;
-    }
+  }
 }
--- a/src/app/forms/validators/custom-async-validators.ts
+++ b/src/app/forms/validators/custom-async-validators.ts
@@ -153,7 +153,7 @@ export class CustomAsyncValidators {
 
     const fieldValue = testResult.testTypes[0].hasOwnProperty(field)
       ? (testResult.testTypes[0] as any)[field]
-      : (testResult as any)[field as keyof TestType];
+      : (testResult as any)[field as keyof TestResultModel];
 
     const isTrue = Array.isArray(value) ? value.includes(fieldValue) : fieldValue === value;
 
--- a/src/app/forms/validators/custom-async-validators.ts
+++ b/src/app/forms/validators/custom-async-validators.ts
@@ -145,13 +145,7 @@ export class CustomAsyncValidators {
       return CustomAsyncValidators.checkCondition(testResult, conditions);
     }
 
-    for (const condition of conditions) {
-      if(!CustomAsyncValidators.checkCondition(testResult, condition)) {
-        return false;
-      }
-    }
-    
-    return true;
+    return conditions.every(condition => CustomAsyncValidators.checkCondition(testResult,condition))
   }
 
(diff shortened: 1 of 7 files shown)


src/app/forms/validators/custom-async-validators.ts
```diff
@@ -114,7 +117,62 @@ export class CustomAsyncValidators {
...
+          if(testResult){
+            const conditionsPassed = CustomAsyncValidators.checkConditions(testResult, conditions);
+
+            if (control?.parent) {
+              const siblingControl = control.parent.get(sibling) as CustomFormControl;
+              siblingControl.meta.hide = Array.isArray(value) ? (conditionsPassed && value.includes(control.value)) : (conditionsPassed && control.value === value);
+            }
+          }
+
+          return null;
```
tahayazici, line 140: This can be clarified.
```suggestion
          if (!testResult || !control?.parent) { return null; }

          const siblingControl = control.parent.get(sibling) as CustomFormControl;

          const conditionsPassed = CustomAsyncValidators.checkConditions(testResult, conditions);

          siblingControl.meta.hide = conditionsPassed && (Array.isArray(value) ? value.includes(control.value) : control.value === value);
```

src/app/forms/validators/custom-async-validators.ts
```diff
@@ -114,7 +117,62 @@ export class CustomAsyncValidators {
...
+      let failed: boolean = false;
+
+      conditions.forEach((condition) => {
+        if(!CustomAsyncValidators.checkCondition(testResult, condition)) failed = true;
+      });
+
+      return !failed;
+    }else{
+      return CustomAsyncValidators.checkCondition(testResult, conditions);
+    }
```
tahayazici, line 156: This can be simplified.
```suggestion
    if (!Array.isArray(conditions)) {
      return CustomAsyncValidators.checkCondition(testResult, conditions);
    }

    for (const condition of conditions) {
      if(!CustomAsyncValidators.checkCondition(testResult, condition)) {
        return false;
      }
    }
    
    return true;
```

src/app/forms/validators/custom-async-validators.ts
```diff
@@ -114,7 +117,62 @@ export class CustomAsyncValidators {
...
+    }else if(testResult?.hasOwnProperty(field)) {
+      fieldValue = (testResult as any)[field];
+    }
+
+    if(operator === operatorEnum.Equals){
+        return Array.isArray(value) ? value.includes(fieldValue) : fieldValue === value
+      }else if(operator === operatorEnum.NotEquals){
+        return Array.isArray(value) ? !value.includes(fieldValue) : fieldValue !== value
+      }
+      return false;
```
tahayazici, line 176: This can be shortened.
```suggestion
    const {field, operator, value} = condition

    const fieldValue = testResult.testTypes[0].hasOwnProperty(field)
      ? (testResult?.testTypes[0] as any)[field]
      : (testResult as any)[field];

    const isTrue = Array.isArray(value) ? value.includes(fieldValue) : fieldValue === value;

    return operator === operatorEnum.Equals ? isTrue : !isTrue;
```
You don't need to worry about whether `testResult?.hasOwnProperty(field)` fails or not, because if it does, `fieldValue` will be undefined anyway and fail the lower test.

tahayazici (changes requested):
Very nice validator. Left some refactoring comments that should be easy to implement.

src/app/forms/validators/custom-async-validators.ts
CormacBJSS: Thanks Taha that looks cleaner I will commit that.

src/app/forms/validators/custom-async-validators.ts
CormacBJSS: Thanks Taha that looks simpler I will commit that.

kpearmanbjss (dismissed):

tahayazici (approved):

Ruairidh-BJSS (approved):

src/app/forms/validators/custom-async-validators.ts
```diff
@@ -114,7 +117,52 @@ export class CustomAsyncValidators {
...
+      return CustomAsyncValidators.checkCondition(testResult, conditions);
+    }
+
+    for (const condition of conditions) {
+      if(!CustomAsyncValidators.checkCondition(testResult, condition)) {
+        return false;
+      }
+    }
+    
+    return true;
```
gjulien-bjss, line 154: I think this is the same logic?
```suggestion
    return conditions.every(condition => CustomAsyncValidators.checkCondition(testResult,condition))

Comment (tom-bjss) [reply src/app/forms/validators/custom-async-validators.ts:154]: Agreed

Comment (CormacBJSS) [reply src/app/forms/validators/custom-async-validators.ts:154]: Thanks Guillaume that looks good. I tested it locally and all good so I will commit that.

File: src/app/forms/validators/custom-async-validators.ts [hunk 5]
```diff
@@ -114,7 +117,52 @@ export class CustomAsyncValidators {
...
+  }
+
+  static checkCondition(testResult: TestResultModel, condition: Condition ){
+    const {field, operator, value} = condition
+
+    const fieldValue = testResult.testTypes[0].hasOwnProperty(field)
+      ? (testResult?.testTypes[0] as any)[field]
```
tom-bjss, line 155: can this be null after you've already checked the condition on line 160 ?

src/app/forms/validators/custom-async-validators.ts
```diff
@@ -114,7 +117,46 @@ export class CustomAsyncValidators {
...
+    }
+
+    return conditions.every(condition => CustomAsyncValidators.checkCondition(testResult,condition))
+  }
+
+  static checkCondition(testResult: TestResultModel, condition: Condition ){
+    const {field, operator, value} = condition
+
+    const fieldValue = testResult.testTypes[0].hasOwnProperty(field)
+      ? (testResult?.testTypes[0] as any)[field]
```
gjulien-bjss, line 155:
```suggestion
      ? (testResult?.testTypes[0])[field as keyof TestType]
```

src/app/forms/validators/custom-async-validators.ts
```diff
@@ -114,7 +117,46 @@ export class CustomAsyncValidators {
...
+      : (testResult as any)[field];
```
gjulien-bjss, line 156:
```suggestion
      : (testResult)[field as keyof TestResultModel];
```

src/app/forms/validators/custom-async-validators.ts
CormacBJSS: Hi Tom good point I will delete the question mark after testResult as I guess it can't be null if line 160 succeeds. Thanks

src/app/forms/validators/custom-async-validators.ts
CormacBJSS: I will leave this as is as it is difficult to get the change to work fully.

src/app/forms/validators/custom-async-validators.ts
CormacBJSS: This change has been made.

src/app/forms/validators/custom-async-validators.ts
CormacBJSS: Thanks this change has been made.

gjulien-bjss (approved):

Ruairidh-BJSS (approved):

PR #646, opened 2022-11-07
fix(cb2-6248): set optional booleans to undefined as default

aaron-downing:
- `alterationMarker` and `departmentalVehicleMaker` are optional `boolean` fields, if an option for them is not selected for them via the radio button input, then their value should be set to `undefined` and not `''` as the backend is expecting a `boolean` value if either of these attributes are defined.
- `alterationMarker` was missing from PSV and has now been added

Before:
image

After:
image


naathanbrown (approved):

naathanbrown (approved):

gjulien-bjss (approved):
Does the backend accept `undefined`? I thought the "trend" was to accept `null` instead

tahayazici (approved):

tom-bjss (approved):

PR #652, opened 2022-11-07
fix(cb2-6642): check the right weights property

tahayazici:
## Can't save new values of Brakes due to previously displayed section locked and half locked values

It turns out that the form event coming from the weights section is not encapsulated by `weights` object like the other sections. This is a quick fix to address that.

[CB2-6642](https://dvsa.atlassian.net/browse/CB2-6642)


naathanbrown (approved):

tom-bjss (approved):
I triple checked them this time :D

PR #654, opened 2022-11-07
feat(cb2-6468): fetch reference data

gjulien-bjss:
## Fetch reference data from api

_Fetch reference data from api_
[CB2-6468](https://dvsa.atlassian.net/browse/CB2-6468)

## Checklist


gjulien-bjss (commented):
No merging before end of sprint/until CORS && Api issues are solved pls, if not everything will break

src/app/services/reference-data/reference-data.service.spec.ts
```diff
@@ -22,66 +36,95 @@ describe('TechnicalRecordService', () => {
...
+          expect(response).toEqual(apiResponse);
         });
+
+        const req = controller.expectOne('https://url/api/v1/reference/COUNTRY_OF_REGISTRATION');
+        req.flush(apiResponse);
       });
     });
 
-    it('should thrown an error if resource type is not given', (done) => {
+    it('should thrown an error if resource type is not given', done => {
```
tom-bjss, line 50: If we're adding negative testing then theres the resource type not valid

gjulien-bjss: The API returns a 200 with an empty "data" array. So the test is not representative of what would actually happen if you called the API with a key that does not exist. I believe that was the reason behind it.

tahayazici (dismissed):

gjulien-bjss:
> So, you're telling me we've got reference data in place for all those deleted mocks?

Well we should go. It is in the repo for the the reference data. The seed job seems somewhat inconsistent though. Hoping it's a data issue that can easily be resolved

mcadenas-bjss (approved):

kpearmanbjss (dismissed):

kpearmanbjss (approved):

mcadenas-bjss (dismissed):

src/app/store/reference-data/effects/reference-data.effects.ts
```diff
@@ -22,9 +23,17 @@ export class ReferenceDataEffects {
...
+        this.referenceDataService.fetchReferenceData(resourceType, paginationToken).pipe(
+          switchMap(data => {
+            if (isPaginated(data)) {
+              return of(
+                fetchReferenceDataSuccess({ resourceType, payload: data.data as ReferenceDataModelBase[] }),
+                fetchReferenceData({ resourceType, paginationToken: data.paginationToken })
+              );
+            }
+            return of(fetchReferenceDataSuccess({ resourceType, payload: data.data as ReferenceDataModelBase[] }));
+          }),
```
tahayazici, line 36: Ternary here maybe?
```suggestion
          switchMap(data => isPaginated(data)
            ? of(
              fetchReferenceDataSuccess({ resourceType, payload: data.data as ReferenceDataModelBase[] }),
              fetchReferenceData({ resourceType, paginationToken: data.paginationToken })
            )
            : of(fetchReferenceDataSuccess({ resourceType, payload: data.data as ReferenceDataModelBase[] }))
          ),
```

andrewf-bjss (dismissed):

src/app/store/reference-data/effects/reference-data.effects.ts
mcadenas-bjss: Don't look as readable :(

gjulien-bjss: I thought about it, but because its's a sort of long winded expression it feels much less readable.

tahayazici: That is fair.

tahayazici (approved):

PR #666, opened 2022-11-11
fix(cb2-6683): fix amending provisional record

aaron-downing:
Previously going to amend a provisional record caused for the record being amended to be the current record. 

This was an issue with the way the `viewableTechRecord` was being populated which is what populates the `editingTechRecord` state value. Now if `'provisional'` is contained within the last 2 parts of the URL, the provisional record will be selected from state.

Before;
image

After:
image

[link to ticket number](https://dvsa.atlassian.net/browse/CB2-6683)

### DIFF
--- a/src/app/services/technical-record/technical-record.service.ts
+++ b/src/app/services/technical-record/technical-record.service.ts
@@ -143,9 +143,16 @@ export class TechnicalRecordService {
       select(selectRouteNestedParams),
       map(params => {
         const createdAt = params['techCreatedAt'];
-        const isProvisional = this.router.url.split('/').pop() === 'provisional';
+        const lastTwoUrlParts = this.router.url.split('/').slice(-2)
 
-        if (isProvisional) {
+        function isProvisional() {
+          if (lastTwoUrlParts.includes('provisional')) {
+            return true
+          }
+          return false;
+        }
+
+        if (isProvisional()) {
           return vehicleRecord.techRecord.find(record => record.statusCode === StatusCodes.PROVISIONAL)
         }
 
--- a/src/app/services/technical-record/technical-record.service.ts
+++ b/src/app/services/technical-record/technical-record.service.ts
@@ -144,8 +144,7 @@ export class TechnicalRecordService {
       map(params => {
         const createdAt = params['techCreatedAt'];
         const lastTwoUrlParts = this.router.url.split('/').slice(-2)
-
-       const isProvisional = () => lastTwoUrlParts.includes('provisional');
+        const isProvisional = () => lastTwoUrlParts.includes('provisional');
 
         if (isProvisional()) {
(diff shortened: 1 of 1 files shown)


naathanbrown (dismissed):

src/app/services/technical-record/technical-record.service.ts
```diff
@@ -143,9 +143,16 @@ export class TechnicalRecordService {
...
-        const isProvisional = this.router.url.split('/').pop() === 'provisional';
+        const lastTwoUrlParts = this.router.url.split('/').slice(-2)
 
-        if (isProvisional) {
+        function isProvisional() {
+          if (lastTwoUrlParts.includes('provisional')) {
+            return true
+          }
+          return false;
+        }
```
mcadenas-bjss, line 153:
```suggestion
       const isProvisional = () => lastTwoUrlParts.includes('provisional');
```
I think you can simplify the function like this.

mcadenas-bjss (approved):

src/app/services/technical-record/technical-record.service.ts
aaron-downing: I wanted it to specifically look at the last two parts of the URL in case of any other potential side effects, but I couldn't exactly think of any off the top of my head

src/app/services/technical-record/technical-record.service.ts
```diff
@@ -143,9 +143,10 @@ export class TechnicalRecordService {
       select(selectRouteNestedParams),
       map(params => {
         const createdAt = params['techCreatedAt'];
-        const isProvisional = this.router.url.split('/').pop() === 'provisional';
+        const lastTwoUrlParts = this.router.url.split('/').slice(-2)
```
mcadenas-bjss, line 146: This doesn't seem like an exact science. Can this break if the route paths change?

mcadenas-bjss (approved):

src/app/services/technical-record/technical-record.service.ts
```diff
@@ -143,9 +143,10 @@ export class TechnicalRecordService {
...
+        const isProvisional = () => lastTwoUrlParts.includes('provisional');
```
gjulien-bjss, line 147: Does it also need to be a function can it not just be in the if?

gjulien-bjss (approved):

src/app/services/technical-record/technical-record.service.ts
aaron-downing: I'm not sure, in what way do you mean break? As in it would become undefined?

src/app/services/technical-record/technical-record.service.ts
aaron-downing: Potentially, although I wanted to be explicit what the if statement is checking by wrapping the value in a named variable/function

PR #685, opened 2022-11-21
feat(cb2-6764): add role guards to new permission

naathanbrown:
## Front end user permissions for archiving Tech Record

Add new role guards and permissions for archiving to the front end.
[https://dvsa.atlassian.net/browse/CB2-6764](6764)

## Checklist

### DIFF
--- a/src/app/models/roles.enum.ts
+++ b/src/app/models/roles.enum.ts
@@ -3,6 +3,7 @@ export enum Roles {
   TechRecordView = 'CVSFullAccess,TechRecord.View',
   TechRecordCreate = 'CVSFullAccess,TechRecord.Create',
   TechRecordAmend = 'CVSFullAccess,TechRecord.Amend',
+  TechRecordArchive = 'CVSFullAccess,TechRecord.Archive',
   TestResultView = 'CVSFullAccess,TestResult.View',
   TestResultAmend = 'CVSFullAccess,TestResult.Amend',
   TestResultCreateContingency = 'CVSFullAccess,TestResult.CreateContingency'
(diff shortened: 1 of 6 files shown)


src/app/models/roles.enum.ts
```diff
@@ -3,6 +3,7 @@ export enum Roles {
   TechRecordView = 'CVSFullAccess,TechRecord.View',
   TechRecordCreate = 'CVSFullAccess,TechRecord.Create',
   TechRecordAmend = 'CVSFullAccess,TechRecord.Amend',
+  TechRecordArchive = 'CVSFullAccess,TechRecord.Archive',
```
mcadenas-bjss, line 6: I'm adding a new role for desk-based test creation and found that the roles here don't look right. Shouldn't each role have only one "action"? e.g. `TechRecordArchive = 'TechRecord.Archive',`.

When we log in, roles are stored a string array: 
image
So then doing something like:
```
userService.roles$.subscribe(roles => {
   if(roles.includes(Roles. TechRecordArchive)) {
      console.log('I can archive');
   }
)
```
This will not log out because there is no role in state with value `CVSFullAccess,TechRecord.Archive`.

gjulien-bjss (commented):
Failing tests

src/app/models/roles.enum.ts
naathanbrown: They're like this so locally you don't need to worry about roles, the production code looks like what you are thinking. There is an env var if you want to toggle it on or off.

gjulien-bjss (approved):

HazelWilderspin (approved):

PR #764, opened 2022-12-16
feat(cb2-6886): prevent test creation on skeleton tech records

tahayazici:
## Ensure contingency tests can't be captured on skeleton Tech Records in VTM

- Add a completeness tag to the tech record title.
- Remove the redundant create test record button component.
- Display a pop-up when the user tries to create a test on a skeleton tech record.

[CB2-6869](https://dvsa.atlassian.net/browse/CB2-6869)
[CB2-6886](https://dvsa.atlassian.net/browse/CB2-6886)


src/app/features/tech-record/components/vehicle-technical-record/vehicle-technical-record.component.ts
```diff
@@ -106,7 +108,32 @@ export class VehicleTechnicalRecordComponent implements OnInit, AfterViewInit {
...
+      default:
+        return 'Unknown Vehicle Type';
+    }
+  }
+
+  createTest(isComplete?: string): void {
+    if (isComplete) {
+      this.router.navigate(['test-records/create-test/type'], { relativeTo: this.route });
+    } else {
+      confirm(
```
Ian-Hodges, line 128: Would alert() be better? confirm() suggests they have an option to continue or cancel.

tahayazici: Suggestion applied.

src/app/features/tech-record/components/vehicle-technical-record/vehicle-technical-record.component.ts
```diff
@@ -106,7 +108,32 @@ export class VehicleTechnicalRecordComponent implements OnInit, AfterViewInit {
...
+      default:
+        return 'Unknown Vehicle Type';
+    }
+  }
+
+  createTest(isComplete?: string): void {
+    if (isComplete) {
+      this.router.navigate(['test-records/create-test/type'], { relativeTo: this.route });
+    } else {
+      alert(
```
aaron-downing, line 128: Should it be alert or should we just add a red banner box at the top of the page?

aaron-downing: A bit like when you try to submit amends on a tech record when you're missing fields?

tahayazici: It should be an alert, that's what they want. Check the ticket.

aaron-downing: But that SD is from an app, might be worth just asking Matt

tahayazici: Just got confirmation from Matt. We're going ahead with this.

Ian-Hodges (approved):

src/app/features/tech-record/components/vehicle-technical-record/vehicle-technical-record.component.ts
```diff
@@ -106,7 +108,32 @@ export class VehicleTechnicalRecordComponent implements OnInit, AfterViewInit {
...
+      case VehicleTypes.HGV:
+        return techRecord.make && techRecord.model ? `${techRecord.make}-${techRecord.model}` : '';
+      default:
+        return 'Unknown Vehicle Type';
+    }
+  }
+
+  createTest(isComplete?: string): void {
+    if (isComplete) {
+      this.router.navigate(['test-records/create-test/type'], { relativeTo: this.route });
```
aaron-downing, line 126: very minor, but you could use `return this.router.navigate(['test-records/create-test/type'], { relativeTo: this.route });` and not need the else statement

tahayazici: That is semantically wrong or confusing as this method's return type is `void`. That change would only get rid of the `} else {` line, so I'll leave it as is for now.

leifkemp-bjss (approved):

src/app/features/tech-record/components/vehicle-technical-record/vehicle-technical-record.component.html
```diff
@@ -1,31 +1,26 @@
...
+        <app-tech-record-history [currentRecord]="currentTechRecord" [vehicleTechRecord]="vehicleTechRecord"></app-tech-record-history>
 
-        <app-test-record-summary *appRoleRequired="roles.TestResultView" [testRecords]="(records$ | async) || []"> </app-test-record-summary>
+        <app-test-record-summary *appRoleRequired="roles.TestResultView" [testRecords]="(records$ | async) || []"></app-test-record-summary>
 
         <ng-container *ngIf="!isArchived && !isEditing">
-          <app-test-record-create-button *appRoleRequired="[roles.TestResultCreateContingency, roles.TestResultCreateDeskAssesment]">
-          </app-test-record-create-button>
+          <div class="govuk-button-group" *appRoleRequired="[roles.TestResultCreateContingency, roles.TestResultCreateDeskAssesment]">
+            <a id="create-test" class="govuk-button" (click)="createTest(currentTechRecord?.recordCompleteness)">Create a test for this vehicle</a>
```
gjulien-bjss, line 22: This should be a button, or it should have an `href` or `routerLink` attribute.

src/app/features/tech-record/components/tech-record-title/tech-record-title.component.ts
```diff
@@ -41,6 +42,14 @@ export class TechRecordTitleComponent implements OnInit {
     return Roles;
   }
 
+  getCompletenessColor(completeness?: string): 'green' | 'red' {
```
tom-bjss, line 45: We should add this to our own technical debt register, Im sure we do something similar for Test results, good opportunity to refactor in the future 😄

tom-bjss (approved):

src/app/features/tech-record/components/vehicle-technical-record/vehicle-technical-record.component.html
tahayazici: The hover and cursor effect is already handled by the gov styling, so it is not necessary.

src/app/features/tech-record/components/tech-record-title/tech-record-title.component.ts
tahayazici: I do not really get what you mean, but sure, yeah 😊

PR #767, opened 2022-12-19
feat(cb2-6054): cert gen api contract

tom-bjss:
Change to front end to pass in email and name to lambda call


src/app/services/user-service/user-service.ts
```diff
@@ -23,13 +23,13 @@ export class UserService implements OnDestroy {
...
               name,
-              username,
-              idTokenClaims: { oid }
+              idTokenClaims: { oid, preferred_username, email }
             },
             accessToken
           }
         } = result;
-        this.logIn({ name, username, oid, accessToken });
+        let userEmail = email || preferred_username;
```
gjulien-bjss, line 31:
```suggestion
        const userEmail = email || preferred_username;
```

src/app/services/user-service/user-service.ts
```diff
@@ -23,13 +23,13 @@ export class UserService implements OnDestroy {
...
+        this.logIn({ name, userEmail, oid, accessToken });
```
naathanbrown, line 32: Should we be allowing them to use the email to log in? Surely it should stay as user name?

tom-bjss: Its from the token after logging in, both of these fields are emails even user_name but username was populated by the upn which can change and shouldnt be used to identify a user. Where as Open ID allows the optional use of email which we have added and as a fail safe, replicate VTA in using the preferred_username

gjulien-bjss (dismissed):

tahayazici (dismissed):

Ian-Hodges (approved):

naathanbrown (dismissed):

Ian-Hodges (approved):

tahayazici (approved):

PR #784, opened 2023-01-11
fix(cb-7173): brake code prefix appears alongside dropdown

owen-corrigan-bjss:
## CB-7173

https://dvsa.atlassian.net/browse/CB2-7173

I have added display flex to the internal-wrapper class so that when amending a vehicle technical record a PSV vehicle the prefix appears next to the dropdown rather than above it.


tahayazici (approved):

mcadenas-bjss (approved):
🥇

PR #789, opened 2023-01-13
fix(cb2-7200): send the full vehicle object to the put endpoint

gjulien-bjss:
## Updating tech records in VTM turns them into skeleton records

Updating tech records in VTM turns them into skeleton records because VTM doesn't send the whole vehicle object. There also needs to be some formatting done on the Vrms, since the `GET` endpoint returns the vrms formatted in a different way than the `PUT` endpoint expects them in.
[Cb2-7200](https://dvsa.atlassian.net/browse/CB2-7200)

## Checklist


src/app/services/technical-record/technical-record.service.ts
```diff
@@ -64,36 +66,32 @@ export class TechnicalRecordService {
...
+    const body: PutVehicleTechRecordModel & { msUserDetails: { msOid: string | undefined; msUser: string } } = {
+      ...this.formatVrmsForUpdatePayload(vehicleTechRecord),
       msUserDetails: { msOid: user.id, msUser: user.name },
       techRecord: [newTechRecord]
     };
 
     return this.http.put<VehicleTechRecordModel>(url, body, { responseType: 'json' });
   }
 
-  private removeUpdateType(techRecord: TechRecordModel, newTechRecord: TechRecordModel) {
```
gjulien-bjss, line 86: Looked to me like this method was redundant, both functions that use it pass `newTechRecord` as a deep clone of `techRecord`, with minor modifications to the `newTechRecord` object before it gets passed in, however it doesn't seem any of the changed properties have an impact on the value of `newTechRecord.updateType` or `techRecord.updateType`.

src/app/features/tech-record/components/tech-record-change-status/tech-record-change-status.component.ts
```diff
@@ -67,12 +59,12 @@ export class TechRecordChangeStatusComponent implements OnInit {
     return this.isPromotion ? 'Promote' : 'Archive';
   }
 
-  goBack(isAfterSubmussion: boolean = false): void {
```
naathanbrown, line 70: Liked it more with the typo personally

src/app/models/vehicle-tech-record.model.ts
```diff
@@ -9,6 +8,11 @@ export interface VehicleTechRecordModel {
   techRecord: TechRecordModel[];
 }
 
+export interface PutVehicleTechRecordModel extends Omit<VehicleTechRecordModel, 'vrms'> {
```
naathanbrown, line 11: Do you think we will need to be sending these when we get to update VRM?

src/app/services/technical-record/technical-record.service.ts
```diff
@@ -117,6 +115,39 @@ export class TechnicalRecordService {
...
+      )
+      .subscribe(data => {
+        if (data) {
+          data.techRecord[0].reasonForCreation = '';
+          this.updateEditingTechRecord(data);
+        }
+      });
+  }
+
+  updateEditingTechRecord(record: TechRecordModel | VehicleTechRecordModel, resetVehicleAttributes = false): void {
```
naathanbrown, line 132: This is complex, and maybe not 100% clear what we are doing here or why you need it, is it worth a small documenting comment? Genuine question, happy either way

naathanbrown (dismissed):

src/app/models/vehicle-tech-record.model.ts
naathanbrown: Ignore this, saw the bit lower down

leifkemp-bjss (dismissed):

src/app/services/technical-record/technical-record.service.ts
gjulien-bjss: I can add a documenting comment. Basically, I'm using the service to interact with the store, so you can pass either the whole vehicle object or just a tech record. This is to avoid making loads of changes to components e.g. passing the whole vehicle object to 10 components, and just keep using the techRecord with most components interact with. The service is then the one responsible for mapping the given `techRecord` to the vehicleRecord. There is an override that essentially resets the vehicle information to the information in state for the current vehicle, before being editing.

naathanbrown (approved):

mcadenas-bjss (approved):

PR #790, opened 2023-01-16
feature(cb2-5718): change vehicle type to hgv

tahayazici:
## Change Vehicle Type to HGV

- Refactor the tech record change type component.
- Write unit tests for the new component.
- Clean up the components used for related workflows.
- Format the names of the newly created tech record components for consistency.

[CB2-5718](https://dvsa.atlassian.net/browse/CB2-5718)


package.json
```diff
@@ -20,7 +20,7 @@
     "docs:json": "compodoc -p ./tsconfig.json -e json -d .",
     "storybook": "npm run docs:json && storybook dev -p 6006",
     "build-storybook": "npm run docs:json && storybook build",
-    "test": "jest --maxWorkers=3 --colors --coverage",
+    "test": "jest --colors --coverage",
```
tahayazici, line 23: Open to feedback with this, I can revert it if needs be.

With `--maxWorkers=3` set, jest was unable to debug locally saying
_"Both --runInBand and --maxWorkers were specified, but these two options do not make sense together. Which is it?"_

Ian-Hodges: My experience is that it's runs quicker locally with --runInBand.

tom-bjss (approved):

naathanbrown (approved):

PR #809, opened 2023-01-25
feat(cb2-7027): create light vehicle test group 3

tahayazici:
## Create a Light Vehicle Test Group 3

[CB2-7027](https://dvsa.atlassian.net/browse/CB2-7027)


src/app/forms/templates/test-records/section-templates/vehicle/group-3-light-vehicle-section.template.ts
```diff
@@ -0,0 +1,76 @@
...
+    {
+      name: 'trailerId',
+      label: 'Trailer ID',
+      value: '',
+      disabled: true,
+      width: FormNodeWidth.L,
+      type: FormNodeTypes.CONTROL,
+      viewType: FormNodeViewTypes.HIDDEN,
+      editType: FormNodeEditTypes.HIDDEN
+    },
```
gjulien-bjss, line 31: Having the trailer id on here doesn't seem quite right

tahayazici: Removed.

Ian-Hodges (approved):

gjulien-bjss (dismissed):

src/app/forms/templates/test-records/master.template.ts
```diff
@@ -475,6 +476,15 @@ export const masterTpl: Record<VehicleTypes, Partial<Record<keyof typeof TEST_TY
       notes: NotesSection,
       reasonForCreation: reasonForCreationSection
     },
+    testTypesDeskBasedGroup3: {
```
leifkemp-bjss, line 479: This is in the master template, which is for amending tests.
Sections for creating tests go in the create master template.

leifkemp-bjss, line 479: It's also in the LGV section for the master template, and this test group is for Small Trailers
I forget if we confirmed what we were going to do about light trailers or not, but I don't think it should be in this part? (correct me if I'm wrong)

tahayazici: Thanks for pointing that out. I have not done any tests in a long while so I was not aware of the difference. I will address this.

tahayazici: Good point man, after reading the ticket carefully, it looks like this PR is blocked. The Light TRL vehicle type needs to be implemented first for this to work. I will bring it in tomorrow's stand up.

tahayazici: This now sits under `small trl`.

tahayazici: This is now moved into the create master template.

gjulien-bjss (approved):

Ruairidh-BJSS (approved):

kpearmanbjss (approved):

PR #825, opened 2023-01-31
feat(cb2-5698): adjust isunique

leifkemp-bjss:
## CB2-5698: Amend VRM

Quick PR to adjust isUnique, to adjust VRM specific behaviour and allow it to ignore archived VRMs
[CB2-5698](https://dvsa.atlassian.net/browse/CB2-5698)

## Checklist

### DIFF
--- a/src/app/services/technical-record/technical-record.service.ts
+++ b/src/app/services/technical-record/technical-record.service.ts
@@ -174,7 +174,14 @@ export class TechnicalRecordService {
         }
 
         const allTechRecords = vehicleTechRecord.flatMap(record => record.techRecord);
-        return allTechRecords.every(record => record.statusCode === StatusCodes.ARCHIVED);
+        const allTechRecordsArchived = allTechRecords.every(record => record.statusCode === StatusCodes.ARCHIVED);
+
+        if (searchType === SEARCH_TYPES.VRM) {
+          const allVrms = vehicleTechRecord.flatMap(record => record.vrms);
+          const primaryVRMFound = allVrms.some(vrm => vrm.isPrimary && vrm.vrm == valueToCheck);
+          return !primaryVRMFound || (primaryVRMFound && allTechRecordsArchived);
+        }
+        return allTechRecordsArchived;
       }),
       catchError((error: HttpErrorResponse) => {
         return (error.status == 404 && of(true)) || throwError(() => error);
--- a/src/app/services/technical-record/technical-record.service.ts
+++ b/src/app/services/technical-record/technical-record.service.ts
@@ -1,10 +1,18 @@
-import { HttpClient } from '@angular/common/http';
+import { HttpClient, HttpErrorResponse } from '@angular/common/http';
 import { Injectable } from '@angular/core';
 import { Router } from '@angular/router';
-import { PutVehicleTechRecordModel, StatusCodes, TechRecordModel, VehicleTechRecordModel } from '@models/vehicle-tech-record.model';
+import {
+  postNewVehicleModel,
+  PutVehicleTechRecordModel,
+  StatusCodes,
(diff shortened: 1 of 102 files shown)


tom-bjss:
The functionality there is how a vrm is classified as unique, not sure why its changed, please explain.

tahayazici (dismissed):

src/app/services/technical-record/technical-record.service.ts
```diff
@@ -173,13 +173,16 @@ export class TechnicalRecordService {
           return true;
         }
 
+        const allTechRecords = vehicleTechRecord.flatMap(record => record.techRecord);
+        const allTechRecordsArchived = allTechRecords.every(record => record.statusCode === StatusCodes.ARCHIVED);
+        if (allTechRecordsArchived) return true;
```
tom-bjss, line 178: add braces

tom-bjss (approved):

tom-bjss (approved):

src/app/services/technical-record/technical-record.service.ts
```diff
@@ -173,13 +173,18 @@ export class TechnicalRecordService {
...
+        }
+
         if (searchType === SEARCH_TYPES.VRM) {
           const allVrms = vehicleTechRecord.flatMap(record => record.vrms);
           const primaryVRMFound = allVrms.some(vrm => vrm.isPrimary && vrm.vrm == valueToCheck);
           return !primaryVRMFound;
         }
-        const allTechRecords = vehicleTechRecord.flatMap(record => record.techRecord);
-        return allTechRecords.every(record => record.statusCode === StatusCodes.ARCHIVED);
+        return allTechRecordsArchived;
```
tahayazici, line 187:
```typescript
        if (
          vehicleTechRecord
            .flatMap(vehicle => vehicle.techRecord)
            .every(record => record.statusCode === StatusCodes.ARCHIVED)
        ) {
          return true;
        }

        if (searchType === SEARCH_TYPES.VRM) {
          return !vehicleTechRecord
            .flatMap(record => record.vrms)
            .some(vrm => vrm.isPrimary && vrm.vrm === valueToCheck);
        }
        
        return false;
```

Why am I suggesting this change?
1. No need to store things in variables because they are not used anywhere else and the conditional logic is not difficult to understand. All you are doing is flattening the records and either verifying the `statusCode` or the `vrm`.
2. Returning `allTechRecordsArchived` is confusing and incorrect at the end, because you will always return `true` on line 179 if `allTechRecordsArchived === true`. Thus, the end return will always be `false`.
3. No need to early exit by checking `vehicleTechRecord.length === 0` as `.flatMap().every()` will return `true` on empty arrays.

You can verify this by running the following on MDN:
```typescript
const array = [];
console.log(array.flatMap(num => num.yolo).every(num => num.jinx === 'jinx'));
```

tahayazici (changes requested):
Can do with some clarity.

tom-bjss (approved):

src/app/services/technical-record/technical-record.service.ts
leifkemp-bjss: Don't think so, using every would mean it'd only be counted as not unique if all the VRMs it finds are primary and equal the value, as opposed to at least one matching that condition with the some.

tahayazici: I think @mcadenas-bjss wins here, I admit defeat.

leifkemp-bjss: Yeah, but using every suggests that for a VRM to not be unique, **all** of the existing VRMs it finds have to be primary and equal to the VRM we're amending to, as opposed to checking for at least one with some.

PR #830, opened 2023-02-01
refactor(cb2-7456): big tech records refactor

tahayazici:
## Update the Editable Tech Record Logic

- Centralize the logic that verifies all tech record summary forms are valid which was duplicated across the Vehicle Tech Record Component and the Hydrate Component. The validation logic being outside of the summary component was causing bugs as it was using `selectedTechRecord` instead of `editableTechRecord`.
- Simplify the naming conventions in common tech record components

[CB2-7456](https://dvsa.atlassian.net/browse/CB2-7456)

### DIFF
--- a/src/app/features/create/components/hydrate-new-vehicle-record/hydrate-new-vehicle-record.component.ts
+++ b/src/app/features/create/components/hydrate-new-vehicle-record/hydrate-new-vehicle-record.component.ts
@@ -36,6 +36,8 @@ export class HydrateNewVehicleRecordComponent {
   }
 
   handleSubmit() {
+    this.summary?.checkForms();
+
     if (!this.isInvalid) {
       this.store.dispatch(createVehicleRecord());
       this.actions$.pipe(ofType(createVehicleRecordSuccess), take(1)).subscribe(() => this.navigateBack());
--- a/src/app/features/create/components/hydrate-new-vehicle-record/hydrate-new-vehicle-record.component.ts
+++ b/src/app/features/create/components/hydrate-new-vehicle-record/hydrate-new-vehicle-record.component.ts
@@ -1,10 +1,7 @@
-import { Component, OnInit, ViewChild } from '@angular/core';
+import { Component, ViewChild } from '@angular/core';
 import { ActivatedRoute, Router } from '@angular/router';
-import { GlobalError } from '@core/components/global-error/global-error.interface';
 import { GlobalErrorService } from '@core/components/global-error/global-error.service';
-import { DynamicFormService } from '@forms/services/dynamic-form.service';
-import { CustomFormGroup, CustomFormArray } from '@forms/services/dynamic-form.types';
-import { VehicleTechRecordModel, VehicleTypes } from '@models/vehicle-tech-record.model';
+import { VehicleTechRecordModel } from '@models/vehicle-tech-record.model';
 import { Actions, ofType } from '@ngrx/effects';
 import { Store } from '@ngrx/store';
 import { TechnicalRecordService } from '@services/technical-record/technical-record.service';
@@ -20,7 +17,6 @@ import { TechRecordSummaryComponent } from '../../../tech-record/components/tech
 export class HydrateNewVehicleRecordComponent {
   @ViewChild(TechRecordSummaryComponent) summary?: TechRecordSummaryComponent;
   isInvalid: boolean = false;
-  vehicleType?: VehicleTypes;
 
   constructor(
     private actions$: Actions,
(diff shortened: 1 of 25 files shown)


src/app/features/tech-record/components/tech-record-title/tech-record-title.component.ts
```diff
@@ -9,25 +9,25 @@ import { editableTechRecord } from '@store/technical-records';
...
   @Input() hideActions: boolean = false;
 
   currentTechRecord$!: Observable<TechRecordModel | undefined>;
-  queryableRecordActions: string[] = [];
-  vehicleMakeAndModel?: string;
+  queryableActions: string[] = [];
+  vehicleMakeAndModel = '';
 
-  constructor(private route: ActivatedRoute, private router: Router, private technicalRecordService: TechnicalRecordService, private store: Store) {}
+  constructor(private route: ActivatedRoute, private router: Router, private store: Store, private technicalRecordService: TechnicalRecordService) {}
```
naathanbrown, line 25: Interested why these have been changed round?

tahayazici: Alphabetical order

HazelWilderspin:
This code breaks the error handling on create new record, it no longer triggers a red global error at the top of the page when you attempt to submit a new record

src/app/features/tech-record/tech-record.component.html
```diff
@@ -1,17 +1,16 @@
...
-  <app-vehicle-technical-record
-    [vehicleTechRecord]="vehicleTechRecord || undefined"
-    *appRoleRequired="roles.TechRecordView"
-  ></app-vehicle-technical-record>
-</ng-container>
+<ng-container *ngIf="vehicle$ | async as vehicle; else noVehicle">
+  <app-vehicle-technical-record *appRoleRequired="roles.TechRecordView" [vehicle]="vehicle"></app-vehicle-technical-record>
 
-<h1 *ngIf="(vehicleTechRecord$ | async) === undefined">This vehicle is not real, like the no vehicle page we need to create</h1>
+  <ng-template #error let-name>
```
tahayazici, line 4: Is this the global red error @HazelWilderspin?

tahayazici:
Forms were being checked on every edit, but not on submission. This is now fixed @HazelWilderspin, thanks for the input.

src/app/features/tech-record/components/tech-record-summary/tech-record-summary.component.ts
```diff
@@ -68,94 +73,147 @@ export class TechRecordSummaryComponent implements OnInit {
...
-    this.formChange.emit();
+    this.technicalRecordService.updateEditingTechRecord(this.techRecordCalculated);
+
+    this.checkForms();
   }
 
-  generateAxleSpacing(numberOfAxles: number, setOriginalValues: boolean = false, axleSpacingOriginal?: AxleSpacing[]): AxleSpacing[] {
-    let axleSpacing: AxleSpacing[] = [];
+  checkForms(): void {
+    const forms = this.sections?.map(section => section.form).concat(this.customSectionForms);
```
gjulien-bjss, line 157: `this.sections` is of type `QueryList<DynamicFormGroupComponent>`, I'm not sure there is a `?` after `this.sections` here.

Of course, the correct implementation would be to remove the `!` on line 32 and make it `QueryList<DynamicFormGroupComponent> | undefined`, in which case the `?` makes sense. The reason I'm not suggesting using a `!` is because the query are ran before the `ngAfterViewInit` lifecycle hook, meaning it's not always defined, at constructor time for instance.

gjulien-bjss, line 157: Additionally this seems to be an new method, would be nice to have some unit tests, especially since it's being used in other components (nice refactor btw 🥳)

src/app/features/create/components/hydrate-new-vehicle-record/hydrate-new-vehicle-record.component.ts
```diff
@@ -35,50 +31,19 @@ export class HydrateNewVehicleRecordComponent {
     return this.technicalRecordService.editableVehicleTechRecord$.pipe(
       tap(vehicle => {
         if (!vehicle) this.navigateBack();
-        this.vehicleType = vehicle?.techRecord[0].vehicleType;
```
gjulien-bjss, line 38: why has this been removed?

tahayazici: Not needed anymore.

mcadenas-bjss (approved):

tom-bjss (approved):

src/app/features/tech-record/components/tech-record-summary/tech-record-summary.component.ts
tahayazici: Because it is a `ViewChild` and not an `Input`, I did not want to force `!` on it. You can enforce inputs by adding them to the selector but viewchilds do not have this advantage. E.g. if the element is dynamic on the template and it is removed by an `*ngIf` at some point, the viewchild will be `null`.

tahayazici: This PR is needed for another ticket that is waiting is testing. I will write up the unit tests for these in another PR. That's for the pointers man 🙂

PR #835, opened 2023-02-03
fix(cb2-9999): correct body type code letter

tahayazici:
## Add new Body Type values to VTM

[CB2-397](https://dvsa.atlassian.net/browse/CB2-397)


kpearmanbjss (approved):

gjulien-bjss (approved):
Couldn't even wait for the action to add the label and assign you 🤣

PR #850, opened 2023-02-09
fix(cb2-7503):  show only most recent previous VRM

owen-corrigan-bjss:
CB2-7503: 

make it so only one previous VRM shows, and it be the most recent previous VRM


owen-corrigan-bjss:
noticed an issue that needs fixing

PR #858, opened 2023-02-13
fix(cb2-7537): Editing Tech Record and Vehicle Tech Record don't always match

HazelWilderspin:
Editing Tech Record and Vehicle Tech Record don't always match

Cleanup editing tech record on navigate to home, search or search result
[Link](https://dvsa.atlassian.net/browse/CB2-7537)

## Checklist

### DIFF
--- a/src/app/features/tech-record/components/tech-record-history/tech-record-history.component.ts
+++ b/src/app/features/tech-record/components/tech-record-history/tech-record-history.component.ts
@@ -51,6 +51,6 @@ export class TechRecordHistoryComponent {
   }
 
   updateTechRecordInEdit(index: number): void {
-    this.techRecordService.updateEditingTechRecord(this.vehicle.techRecord[index]);
+    this.techRecordService.updateEditingTechRecord(this.vehicle.techRecord[this.pageStart || 0 + index]);
   }
 }
--- a/src/app/features/tech-record/components/tech-record-history/tech-record-history.component.ts
+++ b/src/app/features/tech-record/components/tech-record-history/tech-record-history.component.ts
@@ -1,5 +1,6 @@
 import { ChangeDetectionStrategy, ChangeDetectorRef, Component, Input } from '@angular/core';
 import { StatusCodes, TechRecordModel, VehicleTechRecordModel } from '@models/vehicle-tech-record.model';
+import { TechnicalRecordService } from '@services/technical-record/technical-record.service';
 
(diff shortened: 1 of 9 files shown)


src/app/features/tech-record/create/create-tech-records-routing.module.ts
```diff
@@ -1,6 +1,7 @@
 import { NgModule } from '@angular/core';
 import { RouterModule, Routes } from '@angular/router';
 import { MsalGuard } from '@azure/msal-angular';
+import { CancelEditTechGuard } from '@guards/cancel-edit-tech/cancel-edit-tech.guard';
```
tahayazici, line 4: This can be removed.

tahayazici (commented):
Nice fix 👍 I'm guessing you've done your local testing.

src/app/features/tech-record/create/create-tech-records-routing.module.ts
HazelWilderspin: done, yeah I tested it locally

tahayazici (dismissed):

mcadenas-bjss (dismissed):

tahayazici (dismissed):

tahayazici (dismissed):

src/app/features/tech-record/components/tech-record-history/tech-record-history.component.ts
```diff
@@ -48,4 +49,8 @@ export class TechRecordHistoryComponent {
         return `/tech-records/${this.vehicle.systemNumber}/`;
     }
   }
+
+  updateTechRecordInEdit(index: number): void {
+    this.techRecordService.updateEditingTechRecord(this.vehicle.techRecord[this.pageStart || 0 + index]);
```
mcadenas-bjss, line 54: Should this function not be called with `true` as a second param to make sure we update rather than patch? The function is confusing but it seems like the second param makes it so we dispatch a fresh version of the selected tech record, not merge with the current editing version.

tahayazici: Suggestion applied.

tahayazici (approved):

naathanbrown (approved):

Ruairidh-BJSS (approved):

mcadenas-bjss (approved):

PR #874, opened 2023-02-20
feat(cb2-7603): amend vin

owen-corrigan-bjss:
CB2-7603

Created a service method for amendVin, also created associated action and effect

### DIFF
--- a/src/app/store/technical-records/actions/technical-record-service.actions.ts
+++ b/src/app/store/technical-records/actions/technical-record-service.actions.ts
@@ -45,7 +45,7 @@ export const updateTechRecordsSuccess = createOutcomeAction('updateTechRecords',
 export const updateTechRecordsFailure = createOutcomeAction('updateTechRecords');
 
 export const archiveTechRecord = createAction(`${prefix} archiveTechRecord`, props<{ systemNumber: string; reasonForArchiving: string }>());
-export const archiveTechRecordSuccess = createAction('archiveTechRecord Success');
+export const archiveTechRecordSuccess = createOutcomeAction('archiveTechRecord', true);
 export const archiveTechRecordFailure = createOutcomeAction('archiveTechRecord');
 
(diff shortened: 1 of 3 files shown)


src/app/store/technical-records/actions/technical-record-service.actions.ts
```diff
@@ -52,6 +52,10 @@ export const updateEditingTechRecord = createAction(`${prefix} updateEditingTech
 export const updateEditingTechRecordCancel = createAction(`${prefix} updateEditingTechRecordCancel`);
 
 export const changeVehicleType = createAction(`${prefix} changeVehicleType`, props<{ vehicleType: VehicleTypes }>());
+//not sure on this
+export const amendVin = createAction(`${prefix} amendVin`, props<{ oldVin: string; newVin: string; systemNumber: string }>());
+export const amendVinSuccess = createAction(`amendVin Success`);
```
gjulien-bjss, line 57:
```suggestion
export const amendVinSuccess = createOutcomeAction(`amendVin`, true);
```

src/app/store/technical-records/actions/technical-record-service.actions.ts
```diff
@@ -52,6 +52,10 @@ export const updateEditingTechRecord = createAction(`${prefix} updateEditingTech
...
+export const amendVinFailure = createOutcomeAction(`amendVin`);
```
gjulien-bjss, line 58: Not super important, but you don't need the backticks if you're not doing interpolation

src/app/store/technical-records/actions/technical-record-service.actions.ts
owen-corrigan-bjss: I wasn't sure with this one, As the API just responds with a 200 and ok and not a tech record I'd kept the createAction as createOutcomeSuccess seems to respond with an array of vehicle tech records which I didn't think was correct for this action?

mcadenas-bjss (approved):

Ruairidh-BJSS (approved):

PR #879, opened 2023-02-21
feat/CB2-7603

owen-corrigan-bjss:
CB2-7603:

removed references to old vin as its no longer required and created a test to make sure method being used is correct and to correct URL


Ruairidh-BJSS:
Do we want to be merging this into VTMDEV-1?

PR #908, opened 2023-03-02
CB2-7714: Icons for new vehicle classes

owen-corrigan-bjss:
CB2-7714:

Have added some SVG's for new vehicle categories, need the proper ones from product design


HazelWilderspin:
image
Mmmm, nice use of shape and colour. The lines feel a bit undefined, if they were thicker and the same width all round the contrast wouldn't feel quite so disproportionate and would be in keeping with other such designs

leifkemp-bjss:
I think if you had a drawing glove you'd be a master of the arts
10/10 already, but 11/10 if you get a drawing glove

PR #944, opened 2023-03-16
feat(cb2-7652): implement the test cancellation workflow

tahayazici:
## Cancel a test

- Create a new cancel test component.
- Implement the cancellation service method.
- Update the test result tag to include "Cancelled".

[CB2-7652](https://dvsa.atlassian.net/browse/CB2-7652)


src/app/features/test-records/amend/views/confirm-cancellation/confirm-cancellation.component.scss
gjulien-bjss, line 1: should delete this file

src/app/store/test-records/effects/test-records.effects.ts
```diff
@@ -118,6 +155,41 @@ export class TestResultsEffects {
     )
   );
 
+  /**
+   * Call PUT Test Results API to cancel test result
+   */
+  cancelTestResult$ = createEffect(() =>
```
gjulien-bjss, line 161: it seems strange to me that we need a whole new effect for what is effectively the exact same as an update. Would it now work to subscribe to the value changes of the form, update the editingTestResult and then just dispatch the usual test result update?

src/app/features/test-records/amend/views/confirm-cancellation/confirm-cancellation.component.html
```diff
@@ -0,0 +1,17 @@
+<form [formGroup]="form" (submit)="handleSubmit()">
+  <app-vehicle-header [isEditing]="true" [testResult]="testResult$ | async" [testNumber]="testNumber$ | async"></app-vehicle-header>
+
+  <app-text-area formControlName="reason" name="reason" label="Why do you want to cancel this test?"></app-text-area>
+
+  <button
```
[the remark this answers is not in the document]
tahayazici: This is of `type=submit` which the `<app-button>` cannot replicate.

src/app/features/test-records/amend/views/confirm-cancellation/confirm-cancellation.component.scss
tahayazici: Done

src/app/store/test-records/effects/test-records.effects.ts
tahayazici: An example of your suggestion?

But in hindsight, that would go against the dummy component approach. Components, ideally, shouldn't need to how editingTestResult works and how it links to our usual update flow. They should just need to dispatch an action and don't worry about anything else. I would rather have 2 actions that are similar than put any more responsibility on the components.

owen-corrigan-bjss (dismissed):

src/app/store/test-records/effects/test-records.effects.ts
tahayazici: Suggestion applied.

src/app/features/test-records/components/vehicle-header/vehicle-header.component.ts
```diff
@@ -46,8 +48,25 @@ export class VehicleHeaderComponent {
...
+  }
+
+  getTagType(): TagType {
+    if (this.testResult?.testStatus === TestResultStatus.CANCELLED) {
+      return TagType.YELLOW;
+    }
+
+    switch (this.resultOfTest) {
+      case resultOfTestEnum.pass:
+        return TagType.GREEN;
```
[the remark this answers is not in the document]
tahayazici: You're right, I've already modified `get resultOfTest()`. Suggestion applied.

gjulien-bjss (approved):

leifkemp-bjss (approved):

PR #979, opened 2023-03-29
fix(cb2-7932): change state systems to run only on first change

naathanbrown:
## Ticket title

This adds on changes and makes it only run on inital load. The store now only pipes and changes the editing record when it is not falsy.
[https://dvsa.atlassian.net/browse/CB2-7932](7932)

## Checklist

- [ ] Branch is rebased against the latest develop/common
- [ ] Necessary `id` required prepended with `"test-"` have been checked with automation testers and added
- [ ] Code and UI has been tested manually after the additional changes
- [ ] PR title includes the JIRA ticket number
- [ ] Squashed commits contain the JIRA ticket number
- [ ] Link to the PR added to the repo
- [ ] Delete branch after merge

PR #982, opened 2023-04-04
fix(cb2-7966): changing vehicle type

gjulien-bjss:
## Fix changing vehicle type

_This is not a nice fix. But we need quick fix._

Reverts #979 and fixes the issue it was aiming to solve without breaking changing vehicle type.

## Checklist

### DIFF
--- a/src/app/features/tech-record/components/tech-record-summary/tech-record-summary.component.ts
+++ b/src/app/features/tech-record/components/tech-record-summary/tech-record-summary.component.ts
@@ -3,12 +3,10 @@ import {
   Component,
   EventEmitter,
   Input,
-  OnChanges,
   OnDestroy,
   OnInit,
   Output,
   QueryList,
-  SimpleChanges,
   ViewChild,
   ViewChildren
 } from '@angular/core';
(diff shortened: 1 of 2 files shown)


naathanbrown (approved):

gjulien-bjss:
> @gjulien-bjss - https://dvsa.atlassian.net/browse/CB2-7966 is the related ticket reference.
> 
> Are we sure we want to revert #979 ? I hope we're treading carefully as that was a major production issue...

#979 was meant to be a "nicer" version of this "fix". The changes in #979 are intended to implement the logic in this PR. Reverting it was the fastest and cleanest way of getting this PR ready, I mentioned that's what was happening to facilitate the review process.

I have manually checked happy paths and I can:
- create a vehicle
- amend a vehicle (current and provisional) (without the visual glitch)
- change the vehicle type (current and provisional)
- add a notifiable alteration (without the visual glitch)
- promote a record
- archive a record
- change the vin
- change the vrm

leifkemp-bjss (approved):

src/app/features/tech-record/components/tech-record-summary/tech-record-summary.component.ts
```diff
@@ -72,35 +70,36 @@ export class TechRecordSummaryComponent implements OnInit, OnDestroy, OnChanges
   ) {}
 
   ngOnInit(): void {
+    let isOnInit = true;
```
andrewf-bjss, line 73: As this is a guard to run the logic once, perhaps adding `first` to the name might make it's use clearer - what do you think?

```suggestion
    let isFirstOnInit = true;
```

gjulien-bjss: `onInit` only runs once, it's the logic inside the observable which runs multiple times. The thought is that we want that logic when we are running inside the synchronous function `onInit` and if we're not, we want to run a different logic. 

Since there isn't a "second" or "third" onInit, only one, I don't think it would be qualified as "first". We're either inside the onInit lifecycle hook, or we're not.

PR #987, opened 2023-04-12
test(*): add cypress for local e2e testing

gjulien-bjss:
_Adding cypress to the application will allow local end to end testing and allow fast feedback on any core functionality that might have been broken during refactoring or the adding of new functionality._

This is not aimed at replacing regression testing in https://github.com/dvsa/cvs-auto-vtm , but aimed at providing faster feedback for developers.


cypress/fixtures/example.json
gjulien-bjss, line 1: should delete this file

owen-corrigan-bjss (approved):

PR #1020, opened 2023-05-05
fix(cb2-8192): ref-data admin - adds validation for required field 'resourceKey'

emilyrkelly21:
There is no validation for blank resource key when adding Country of registration

[CB2-8192](https://dvsa.atlassian.net/browse/CB2-8192)

## Checklist

### DIFF
--- a/src/app/features/reference-data/reference-data-add/reference-data-add.component.ts
+++ b/src/app/features/reference-data/reference-data-add/reference-data-add.component.ts
@@ -1,4 +1,4 @@
-import { Component, QueryList, ViewChildren } from '@angular/core';
+import { Component, OnInit, QueryList, ViewChildren } from '@angular/core';
 import { ActivatedRoute, Router } from '@angular/router';
 import { GlobalError } from '@core/components/global-error/global-error.interface';
 import { GlobalErrorService } from '@core/components/global-error/global-error.service';
@@ -27,7 +27,7 @@ import { take } from 'rxjs';
   selector: 'app-reference-data-add',
   templateUrl: './reference-data-add.component.html'
 })
-export class ReferenceDataCreateComponent {
+export class ReferenceDataCreateComponent implements OnInit {
   isEditing: boolean = true;
   type: ReferenceDataResourceType = ReferenceDataResourceType.Brakes;
   newRefData: any;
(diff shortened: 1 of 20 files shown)


owen-corrigan-bjss (dismissed):

andrewf-bjss (dismissed):

src/app/features/reference-data/reference-data-list/reference-data-list.component.ts
```diff
@@ -1,24 +1,37 @@
...
-  constructor(private referenceDataService: ReferenceDataService, private route: ActivatedRoute, private router: Router, private store: Store) {
-    this.referenceDataService.loadReferenceData(ReferenceDataResourceType.ReferenceDataAdminType);
+  constructor(private referenceDataService: ReferenceDataService, private route: ActivatedRoute, private router: Router, private store: Store) {}
 
-    this.route.queryParams.pipe(take(1)).subscribe(params => {
+  ngOnInit(): void {
+    this.route.params.pipe(take(1)).subscribe(params => {
       this.type = params['type'];
+
+      console.log('This is the type by url params: ', this.type);
```
gjulien-bjss, line 34: Probably don't want to `console.log` here 😉

src/app/features/reference-data/reference-data-add/reference-data-add.component.ts
```diff
@@ -39,34 +44,14 @@ export class AddReferenceDataComponent {
...
-                  type: FormNodeTypes.CONTROL,
-                  editType: (referenceData as any)[property]
-                },
-                '',
-                [Validators.required]
-              )
-            )
-          );
-      });
+      console.log('This is the type by url params: ', this.type);
```
gjulien-bjss, line 53: `console.log` probably needs removing

src/app/features/reference-data/reference-data-add/reference-data-add.component.ts
```diff
@@ -77,13 +62,52 @@ export class AddReferenceDataComponent {
...
+      default:
+        templateToReturn = {} as FormNode;
+        break;
+    }
+    templateToReturn.children?.forEach(child => {
+      if (child.name === 'resourceKey') {
+        child.disabled = false;
+      }
+    });
+    return templateToReturn;
```
gjulien-bjss, line 110: I'm not sure why you need to do that if the control is disabled in the controls anyways. If you do that, you can also get rid of the `let` and the `break`s, but just returning inside of the switch statement.

gjulien-bjss, line 110: Additionally, you could return the template which has the `name` equal to the `ReferenceDataResourceType`

owen-corrigan-bjss (approved):

naathanbrown (approved):

PR #1041, opened 2023-06-14
fix(cb2-8358): content change plus vehicle type refactor

shivangidas:
## Batch create summary screen not displaying

1. For the main bug - Using takeUntil in EditTechRecordButtonComponent to unsubscribe from multiple observables
2. Updated the content displayed when no vehicle is found
3. Added the vehicle type to the store - to avoid digging it up from current editingtechrecord
4. Changes to select the current or provisional record for update (only psv and hgv)

[CB2-8358](https://dvsa.atlassian.net/browse/CB2-8358)

## Checklist

### DIFF
--- a/src/app/features/tech-record/create-batch/components/batch-vehicle-results/batch-vehicle-results.component.ts
+++ b/src/app/features/tech-record/create-batch/components/batch-vehicle-results/batch-vehicle-results.component.ts
@@ -1,9 +1,8 @@
 import { Component, OnDestroy } from '@angular/core';
 import { Router, ActivatedRoute } from '@angular/router';
-import { StatusCodes, VehicleTypes } from '@models/vehicle-tech-record.model';
+import { StatusCodes } from '@models/vehicle-tech-record.model';
 import { BatchTechnicalRecordService } from '@services/batch-technical-record/batch-technical-record.service';
 import { TechnicalRecordService } from '@services/technical-record/technical-record.service';
-import { BatchRecord } from '@store/technical-records/reducers/batch-create.reducer';
 import { filter, race, Subject, take, withLatestFrom } from 'rxjs';
 
 @Component({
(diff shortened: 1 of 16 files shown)


src/app/features/tech-record/create-batch/components/batch-vehicle-results/batch-vehicle-results.component.ts
```diff
@@ -11,18 +12,17 @@ import { filter, race, Subject, take, withLatestFrom } from 'rxjs';
 })
 export class BatchVehicleResultsComponent implements OnDestroy {
   private destroy$ = new Subject<void>();
-  techRecord?: TechRecordModel;
+  vehicleRecord?: BatchRecord;
```
gjulien-bjss, line 15: wouldn't call it `vehicleRecord` if it's a batch record

src/app/features/tech-record/create-batch/components/batch-vehicle-results/batch-vehicle-results.component.ts
```diff
@@ -47,7 +47,7 @@ export class BatchVehicleResultsComponent implements OnDestroy {
   }
 
   get vehicleType(): VehicleTypes | undefined {
-    return this.techRecord ? this.technicalRecordService.getVehicleTypeWithSmallTrl(this.techRecord) : undefined;
+    return this.vehicleRecord ? (this.vehicleRecord.vehicleType as VehicleTypes) : undefined;
```
gjulien-bjss, line 50: do we not need the small trailer vehicle type on here ? 

Additionally, if `vehicleRecord` is only used to get the vehicle type, I would probably said to make a public property called `vehicleType` and removing `vehicleRecord` from the component.

Or you can do something like this, if vehicleType is only used in html:
```ts
get vehicleType$(): Observable<VehicleTypes | undefined> {
  this.batchRecordService.batchVehicles$.pipe(map(batchVehicle => batchvehicle?.vehicleType))
}
```
and then `async` pipe it in html

shivangidas: changing this logic to save Vehicle type when selected the first time for batch creation and accessing from that on every page after

kcarr-bjss (dismissed):

naathanbrown (approved):

PR #1053, opened 2023-06-27
feat(cb2-8424): Limit what contingency tests can be created for a Provisional Tech Record

shivangidas:
## Limit what contingency tests can be created for a Provisional Tech Record

[CB2-8424](https://dvsa.atlassian.net/browse/CB2-8424)

## Checklist

### DIFF
--- a/src/app/features/tech-record/components/vehicle-technical-record/vehicle-technical-record.component.ts
+++ b/src/app/features/tech-record/components/vehicle-technical-record/vehicle-technical-record.component.ts
@@ -109,7 +109,11 @@ export class VehicleTechnicalRecordComponent implements OnInit {
   }
 
   showCreateTestButton(vehicleType: VehicleTypes): boolean {
-    return !this.isArchived && !this.isEditing && (this.isCurrent || vehicleType === VehicleTypes.TRL || vehicleType === VehicleTypes.HGV);
+    return (
+      !this.isArchived &&
+      !this.isEditing &&
+      (this.isCurrent || vehicleType === VehicleTypes.TRL || vehicleType === VehicleTypes.HGV || vehicleType === VehicleTypes.PSV)
+    );
   }
 
   createTest(techRecord?: TechRecordModel): void {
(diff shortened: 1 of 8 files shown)


kpearmanbjss (approved):

src/app/features/tech-record/components/vehicle-technical-record/vehicle-technical-record.component.ts
```diff
@@ -108,6 +108,14 @@ export class VehicleTechnicalRecordComponent implements OnInit {
     }
   }
 
+  showCreateTestButton(vehicleType: VehicleTypes): boolean {
+    return (
+      !this.isArchived &&
+      !this.isEditing &&
+      (this.isCurrent || vehicleType === VehicleTypes.TRL || vehicleType === VehicleTypes.HGV || vehicleType === VehicleTypes.PSV)
```
gjulien-bjss, line 115: Is this to remove the "create test button" on provisional records for LGVs, Motorcycles and Cars?

I would say to not bake in some logic to prevent users from creating tests on provisional LGV etc records. Just in case we would want to add those later. Worst case is users just get shown a page with no tests on available on them.

shivangidas: Yup, it is to remove the button.
Can get this looked at tomorrow, need to discuss the PSV change too

shivangidas: We do want to hide the button for now, because otherwise, the specialist tests, technical tests, etc., options will show up for LGVs too as they are in the same parent JSON object as HGV and TRL. The last selection option won't show up as it wouldn't satisfy the provisional condition so the user won't be able to create a test but at this point best to just hide the button for vehicles that don't support creating a test on a provisional record.

gjulien-bjss (approved):

owen-corrigan-bjss (approved):

PR #1075, opened 2023-07-18
feat(cb2-8712): validate on blur

Daniel-Searle:
## Date error messages display too early

This PR is to prevent date validation from triggering until the entire date has been entered.
[CB2-8712](https://dvsa.atlassian.net/browse/CB2-8712)

## Checklist

### DIFF
--- a/src/app/forms/components/date/date.component.html
+++ b/src/app/forms/components/date/date.component.html
@@ -3,7 +3,7 @@
     <legend class="govuk-fieldset__legend govuk-fieldset__legend--m">{{ label }}</legend>
     <div *ngIf="hint" id="{{ name }}-hint" class="govuk-hint">{{ hint }}</div>
     <app-field-error-message
-      *ngIf="dayEl.dirty && monthEl.dirty && yearEl.dirty && yearEl.valid"
+      *ngIf="dayEl.dirty && monthEl.dirty && yearEl.dirty && (yearEl.value ?? '' > 3)"
       [error]="error"
       [name]="name"
     ></app-field-error-message>
@@ -17,7 +17,9 @@
             appFocusNext
             [displayTime]="displayTime"
             class="govuk-input govuk-date-input__input govuk-input--width-2"
-            [ngClass]="{ 'govuk-input--error': elementHasErrors(0) && dayEl.dirty && monthEl.dirty && yearEl.dirty && yearEl.valid }"
+            [ngClass]="{
+              'govuk-input--error': elementHasErrors(0) && dayEl.dirty && monthEl.dirty && yearEl.dirty && (yearEl.value ?? '').length > 3
+            }"
             id="{{ name }}-day"
             name="{{ name }}-day"
             type="number"
@@ -40,7 +42,9 @@
             appFocusNext
             [displayTime]="displayTime"
             class="govuk-input govuk-date-input__input govuk-input--width-2"
-            [ngClass]="{ 'govuk-input--error': dayEl.dirty && monthEl.dirty && yearEl.dirty && yearEl.valid && elementHasErrors(1) }"
+            [ngClass]="{
+              'govuk-input--error': dayEl.dirty && monthEl.dirty && yearEl.dirty && (yearEl.value ?? '').length > 3 && elementHasErrors(1)
+            }"
             id="{{ name }}-month"
             name="{{ name }}-month"
             type="number"
@@ -64,7 +68,8 @@
             [displayTime]="displayTime"
             class="govuk-input govuk-date-input__input govuk-input--width-4"
             [ngClass]="{
-              'govuk-input--error': dayEl.dirty && monthEl.dirty && yearEl.dirty && yearEl.valid && (elementHasErrors(2) || yearEl.invalid)
+              'govuk-input--error':
+                dayEl.dirty && monthEl.dirty && yearEl.dirty && (yearEl.value ?? '').length > 3 && (elementHasErrors(2) || yearEl.invalid)
             }"
             id="{{ name }}-year"
             name="{{ name }}-year"
@@ -95,7 +100,7 @@
                   dayEl.dirty &&
                   monthEl.dirty &&
                   yearEl.dirty &&
-                  yearEl.valid &&
+                  (yearEl.value ?? '').length > 3 &&
                   displayTime &&
                   value &&
                   (undefined === hour || null === hour || hourEl.invalid)
@@ -127,7 +132,7 @@
                   dayEl.dirty &&
                   monthEl.dirty &&
                   yearEl.dirty &&
-                  yearEl.valid &&
+                  (yearEl.value ?? '').length > 3 &&
                   displayTime &&
                   value &&
                   (undefined === minute || null === minute || minuteEl.invalid)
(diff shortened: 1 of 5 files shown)


src/app/forms/components/date/date.component.ts
```diff
@@ -72,7 +72,7 @@ export class DateComponent extends BaseControlComponent implements OnInit, OnDes
       minutes: this.originalDate ? new Date(this.originalDate).getMinutes() : '00',
       seconds: this.originalDate ? new Date(this.originalDate).getSeconds() : '00'
     };
-    this.addValidators();
+    // this.addValidators();
```
naathanbrown, line 75: remove it if we don't need it

gjulien-bjss (commented):
what happens if the users fill in the box, doesn't blur out, and then clicks submit? (let's say on a test for example?) It seems like we're only adding the validator on blur, so I'm thinking it might let you submit invalid dates.

src/app/forms/validators/date/date.validators.ts
```diff
@@ -6,9 +6,24 @@ export class DateValidators {
...
+        return null;
+      }
+      if (!year) {
+        return null;
+      }
+
+      if (year.length < 4) {
+        return null;
+      }
+
```
gjulien-bjss, line 26: I don't think adding these is the way to go. the responsibility of the validator is the validate a date, it should say if a given input is a valid date or not. Handling the displaying or not of errors should not be the validator's responsibility

Daniel-Searle:
> what happens if the users fill in the box, doesn't blur out, and then clicks submit? (let's say on a test for example?) It seems like we're only adding the validator on blur, so I'm thinking it might let you submit invalid dates.

When they go to press submit, surely it'll blur no as they've clicked out of it?

src/app/forms/components/date/date.component.ts
```diff
@@ -158,7 +157,22 @@ export class DateComponent extends BaseControlComponent implements OnInit, OnDes
...
+    }
+    if (!this.year) {
+      return;
+    }
+
+    if (this.year && event.type !== 'blur') {
+      return;
+    }
     this.errors = validateDate(this.day || '', this.month || '', this.year || '', this.label);
   }
```
gjulien-bjss, line 178: I would suggest handling the displaying or not of errors in the `elementHasErrors` method below. We ideally don't want to stop validators running or anything like that.

Additionally you could toggle the visibility of the `<app-field-error-message` component in the `html` of the component using some logic based in the `ts` of the component. Something along the lines of all the elements being touched for instance.

src/app/forms/validators/date/date.validators.ts
Daniel-Searle: > I don't think adding these if statements is the way to go. the responsibility of the validator is the validate a date, it should say if a given input is a valid date or not. Handling the displaying or not of errors should not be the validator's responsibility

Where would you suggest this logic to happen?

gjulien-bjss:
> > what happens if the users fill in the box, doesn't blur out, and then clicks submit? (let's say on a test for example?) It seems like we're only adding the validator on blur, so I'm thinking it might let you submit invalid dates.
> 
> When they go to press submit, surely it'll blur no as they've clicked out of it?

Nope, they don't run, because clicking on another element doesn't blur out of the other one, basically
image

naathanbrown:
> what happens if the users fill in the box, doesn't blur out, and then clicks submit? (let's say on a test for example?) It seems like we're only adding the validator on blur, so I'm thinking it might let you submit invalid dates.

This is true but it does raise the question how is it possible to do this then? We need some way of knowing the user has touched and left the year box?

src/app/forms/components/date/date.component.ts
Daniel-Searle: It seems like the function elementHasErrors doesn't have access to the event, not sure how I could access the event type otherwise? As I can't pass in the event into it

elementHasErrors(i: number, event: Event) --> produces an error in the html

naathanbrown (dismissed):

naathanbrown (dismissed):

src/app/forms/components/date/date.component.html
```diff
@@ -2,7 +2,11 @@
   <fieldset class="govuk-fieldset" role="group" attr.aria-describedby="{{ name }}-hint">
     <legend class="govuk-fieldset__legend govuk-fieldset__legend--m">{{ label }}</legend>
     <div *ngIf="hint" id="{{ name }}-hint" class="govuk-hint">{{ hint }}</div>
-    <app-field-error-message [error]="error" [name]="name"></app-field-error-message>
+    <app-field-error-message
+      *ngIf="dayEl.dirty && monthEl.dirty && yearEl.dirty && yearEl.valid"
```
gjulien-bjss, line 6: why `yearEl.valid` ? don't we want to show an error in the year is not valid?

gjulien-bjss: looks good otherwise

gjulien-bjss: isn't just checking all of them are `dirty` enough?

naathanbrown: No, because as soon as you push '1' into year you get the error message again

naathanbrown: it really needs to be year dirty but not being touched, not sure how to do that here?

gjulien-bjss: makes sense... I personally would say that would be fine, but it's not really up to me...

I that case, I'd say slight nitpick: can we check the length is greater than 3 then ?

naathanbrown: I like checking year.length > 3 as well yeah! I'll see if I can work out a way to tell if the control is currently being touched

naathanbrown: I think the sell on this is 'perfection' is probably really hard and just covering off the likely scenario, plus having it validate 100% at submission is probably okay

naathanbrown:
```suggestion
      *ngIf="dayEl.dirty && monthEl.dirty && yearEl.dirty && yearEl.value.length > 3"
```

naathanbrown (approved):

src/app/forms/components/date/date.component.html
```diff
@@ -2,7 +2,11 @@
   <fieldset class="govuk-fieldset" role="group" attr.aria-describedby="{{ name }}-hint">
     <legend class="govuk-fieldset__legend govuk-fieldset__legend--m">{{ label }}</legend>
     <div *ngIf="hint" id="{{ name }}-hint" class="govuk-hint">{{ hint }}</div>
-    <app-field-error-message [error]="error" [name]="name"></app-field-error-message>
+    <app-field-error-message
+      *ngIf="(dayEl.dirty && monthEl.dirty && yearEl.dirty && (yearEl.value ?? '').length > 3) || displayTime"
```
naathanbrown, line 6:
```suggestion
      *ngIf="(dayEl.dirty && monthEl.dirty && yearEl.dirty && (yearEl.value ?? '').length > 3) && !displayTime"
```

src/app/forms/components/date/date.component.html
```diff
@@ -75,6 +86,11 @@
       </div>
 
       <ng-container *ngIf="displayTime">
+        <app-field-error-message
+          *ngIf="(dayEl.dirty && monthEl.dirty && yearEl.dirty && (yearEl.value ?? '').length > 3) || (displayTime && hourEl.dirty)"
```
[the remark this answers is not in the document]
naathanbrown: infact should be greater than 0

naathanbrown (approved):

shivangidas (approved):

PR #1082, opened 2023-07-20
fix(cb2-8764): implements use of breadcrumbs within reference data administration

emilyrkelly21:
Ref Data: Breadcrumb navigation isn't correct

Ref Data: Breadcrumb navigation isn't correct
[CB2-8764](https://dvsa.atlassian.net/browse/CB2-8764)

## Checklist

### DIFF
--- a/src/app/features/reference-data/reference-data-amend/reference-data-amend.component.ts
+++ b/src/app/features/reference-data/reference-data-amend/reference-data-amend.component.ts
@@ -35,13 +35,17 @@ export class ReferenceDataAmendComponent implements OnInit {
   ) {}
 
   ngOnInit(): void {
-    this.route.params.pipe(first()).subscribe(params => {
+    this.route.parent?.params.pipe(first()).subscribe(params => {
       this.type = params['type'];
+      // load the reference data admin type
+      this.referenceDataService.loadReferenceDataByKey(ReferenceDataResourceType.ReferenceDataAdminType, this.type);
+    });
+
+    this.route.params.pipe(first()).subscribe(params => {
       this.key = decodeURIComponent(params['key']);
 
       if (this.type && this.key) {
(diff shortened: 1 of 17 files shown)


src/app/features/reference-data/reference-data-amend/reference-data-amend.component.ts
```diff
@@ -35,13 +35,17 @@ export class ReferenceDataAmendComponent implements OnInit {
...
   ngOnInit(): void {
-    this.route.params.pipe(first()).subscribe(params => {
+    this.route.parent?.params.pipe(first()).subscribe(params => {
       this.type = params['type'];
+      // load the reference data admin type
+      this.referenceDataService.loadReferenceDataByKey(ReferenceDataResourceType.ReferenceDataAdminType, this.type);
+    });
+
+    this.route.params.pipe(first()).subscribe(params => {
       this.key = decodeURIComponent(params['key']);
```
andrewf-bjss, line 45: Does the parent route params contain the `key` as well?  I'm wondering if it's possible to combine these two very similar if statements to get the `type` and `key` from the same `params`

andrewf-bjss (dismissed):

src/app/features/reference-data/reference-data-amend/reference-data-amend.component.ts
emilyrkelly21: You're presented with a 404 when trying to get them both from the parent route

andrewf-bjss: Can we get them both from the current (child) route, or are we happy with this approach?

gjulien-bjss (approved):

owen-corrigan-bjss (approved):

PR #1089, opened 2023-07-26
feat(cb2-8776): show error message not showing above field

Daniel-Searle:
## Fix error message not showing above field

This PR is prevent submission when test record time is missing
[CB2-8776](https://dvsa.atlassian.net/browse/CB2-8776)

## Checklist

### DIFF
--- a/src/app/forms/validators/date/date.validators.ts
+++ b/src/app/forms/validators/date/date.validators.ts
@@ -20,7 +20,6 @@ export class DateValidators {
       }
 
       if (displayTime) {
-        console.log(t);
         return this.validateTime(t, label);
       }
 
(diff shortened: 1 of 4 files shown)


kcarr-bjss (dismissed):

src/app/forms/validators/date/date.validators.ts
```diff
@@ -29,8 +29,7 @@ export class DateValidators {
 
   private static validateTime(time: string, label: string | undefined) {
     const [hours, minutes] = time.split(':');
-
-    if (!hours || !minutes) {
+    if (hours === '00' || minutes == '00') {
```
naathanbrown, line 32: what if the test is done at 00:00? That is a valid time no?

kpearmanbjss: Is this not also implying you can't type 00 in the minutes box at all meaning all tests that are conducted on the hour will be invalid times?

kcarr-bjss (approved):

naathanbrown (approved):

emilyrkelly21 (approved):

PR #1125, opened 2023-08-30
chore(cb2-8667): update mock tech records

owen-corrigan-bjss:
chore: fixes mockTechRecord functions so they only have information relevant to vehicle type. Removes some mocks from tests in favour of these methods.


naathanbrown (approved):

gjulien-bjss (commented):
The reason most tests don't use the mock tech record is because if you change the data in the mock tech record, tests break. You end up having a dependency between mock data and like 50% of your tests, which is not a good thing. This was especially a problem at the beginning of the project when we used the mock objects to spin up a mock API. 

I don't think migrating more tests to use the mockTechRecord is right way to do it. Tests should supply their own data where possible.

mattwaldronbjss (approved):

shivangidas (dismissed):

PR #1144, opened 2023-09-12
feat(cb2 8225): Unarchive Technical Record

kcarr-bjss:
## Ticket title

Adds functionality to resurrect/unarchive an archived record where there are no other records which share the same VRM which are not archived.
[CB2-8225](https://dvsa.atlassian.net/browse/CB2-8225)

## Checklist


src/app/features/tech-record/tech-record-routing.module.ts
```diff
@@ -48,6 +49,13 @@ const routes: Routes = [
     canActivate: [MsalGuard, CancelEditTechGuard],
     resolve: { load: TechRecordViewResolver }
   },
+  {
+    path: 'historic/unarchive-record',
```
naathanbrown, line 53: if we have the time looking at the router and working out /historic would be great

PR #1160, opened 2023-09-19
poc(cb2-9046): warning validator system

gjulien-bjss:
## Ticket title

this is a POC on how to implement a warning system on the custom controls.

This leverages Angular's validators system by mimicking that logic, to generate warnings stored against the controls.

It is then possible to get all the warnings and add them to state, similar to how is being done for the error system.

if you desire, checkout the branch, and you'll see you get a log in the console if psv notes contains the word "foobar". This will only be triggered upon rendering of the technical record component, so will not be triggered on amendment of psv notes. this is by design, it could be triggered anytime.

## Checklist

### DIFF
--- a/src/app/forms/services/dynamic-form.types.ts
+++ b/src/app/forms/services/dynamic-form.types.ts
@@ -7,19 +7,19 @@ import {
   FormControl,
   FormControlOptions,
   FormGroup,
+  ValidationErrors,
   ValidatorFn
 } from '@angular/forms';
+import { Params } from '@angular/router';
 import { AsyncValidatorNames } from '@forms/models/async-validators.enum';
 import { ValidatorNames } from '@forms/models/validators.enum';
 import { ReferenceDataResourceType } from '@models/reference-data.model';
 import { Store } from '@ngrx/store';
+import { TagType } from '@shared/components/tag/tag.component';
 import { State } from '@store/.';
-import { map, Observable } from 'rxjs';
+import { Observable, firstValueFrom, map } from 'rxjs';
 import { DynamicFormService } from './dynamic-form.service';
 import { SpecialRefData } from './multi-options.service';
-import { Params } from '@angular/router';
-import { TagType } from '@shared/components/tag/tag.component';
-import { Dictionary } from '@ngrx/entity';
 
 export enum FormNodeViewTypes {
   DATE = 'date',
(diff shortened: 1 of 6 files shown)


src/app/forms/services/dynamic-form.types.ts
```diff
@@ -160,6 +164,44 @@ export class CustomFormControl extends FormControl implements CustomControl {
...
+      }
+    });
+
+    if (!errors.length) {
+      this.meta.warning = null;
+    } else {
+      this.meta.warning = errors;
+    }
+  }
+}
```
gjulien-bjss, line 199: These methods are repeated 3 times, once in the `CustomFormControl`, `CustomFormArray` and `CustomFormGroup`. This is mostly because classes can only extend one class.

There is a way using `bind` to reduce this duplication of code. We have used this successfully for `getCleanValue` function in this file.

src/app/forms/templates/psv/psv-notes.template.ts
```diff
@@ -12,7 +13,17 @@ export const PsvNotes: FormNode = {
...
+
+      // Obviously we don't want to add functions in the JSON templates, but it could be handled using a map, the same way other validators are handled
+      warningValidators: [
+        (control: AbstractControl): ValidationErrors | null => {
+          if (control.value.includes('foobar')) {
+            return { test: 'should not equal foobar' };
+          }
+          return null;
+        }
+      ]
```
gjulien-bjss, line 26: then we just add warning validators the exact same way we add normal validators, all through templates.

src/app/forms/services/dynamic-form.types.ts
```diff
@@ -111,8 +111,10 @@ export interface FormNode {
   isoDate?: boolean;
   class?: string;
   customId?: string;
-  warning?: string;
+  warning?: ValidationErrors | null;
```
gjulien-bjss, line 114: we might want to move the warning from being inside `meta` to just being a public property on the classes below. There isn't really a need to add it to the `FormNode` interface, since there is no need for it to be passed in trough templates. This would also stop erroneously passing it trough templates.

src/app/forms/services/dynamic-form.types.ts
```diff
@@ -160,6 +164,44 @@ export class CustomFormControl extends FormControl implements CustomControl {
     super(formState, validatorOrOpts, asyncValidator);
     this.meta = meta;
   }
+
+  addWarningValidators(fn: ValidatorFn) {
+    this.warningValidators = [...this.warningValidators, fn];
```
gjulien-bjss, line 169: can probably `.push` instead of spreading... should be more memory efficient.

PR #1171, opened 2023-09-21
feat(cb2-7464): plate generation validation

kpearmanbjss:
Plate Generation Validation

Enables validation for generating plates and places tags on the relevant fields
[CB2-7464](https://dvsa.atlassian.net/browse/CB2-7464)

## Checklist


src/app/forms/custom-sections/plates/plates.component.ts
```diff
@@ -112,4 +126,35 @@ export class PlatesComponent implements OnInit, OnDestroy, OnChanges {
     }
     return '';
   }
+
+  validateTechRecordPlates(): void {
+    this.globalErrorService.clearErrors();
+    const plateFieldsErrorMessage = 'All fields marked plate are mandatory to generate a plate.';
```
naathanbrown, line 132: Does this need to be a const? We only use it once right? And does it have the correct scope? I can only see it being used in this if block below.

naathanbrown (dismissed):
Two tiny nit picks, really nice PR!

src/app/forms/custom-sections/plates/plates.component.ts
```diff
@@ -112,4 +126,35 @@ export class PlatesComponent implements OnInit, OnDestroy, OnChanges {
...
+    const plateValidationTable = this.techRecord.techRecord_vehicleType === 'trl' ? trlRequiredFields : hgvRequiredFields;
```
Daniel-Searle, line 133: Could maybe rename this to required fields, something like that

src/app/forms/custom-sections/plates/plates.component.ts
```diff
@@ -112,4 +126,35 @@ export class PlatesComponent implements OnInit, OnDestroy, OnChanges {
...
+    if (this.cannotGeneratePlate(plateValidationTable)) {
+      this.viewportScroller.scrollToPosition([0, 0]);
+      this.globalErrorService.addError({ error: plateFieldsErrorMessage });
+      return;
+    }
+    this.store.dispatch(canGeneratePlate());
+    this.router.navigate(['generate-plate'], { relativeTo: this.route });
+  }
+
+  private cannotGeneratePlate(plateValidationTable: string[]) {
```
Daniel-Searle, line 144: I think this method could be refined a bit more, also think maybe the method could be renamed

src/app/forms/custom-sections/plates/platesInterfaces.ts
```diff
@@ -0,0 +1,25 @@
+export interface hgvRequiredPlateFields {
```
Daniel-Searle, line 1: This aint being used no where

src/app/forms/custom-sections/plates/plates.component.ts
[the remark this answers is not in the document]
kpearmanbjss: Could rename it to list instead of table if that makes more sense?

src/app/forms/custom-sections/plates/plates.component.ts
kpearmanbjss: We went with the name because on the if statement it reads clearly as "If this cannot generate plate" but i couldn't think of a better alternative name for it

src/app/forms/custom-sections/plates/platesInterfaces.ts
kpearmanbjss, line 1: I think this was left over from an earlier pr when working on the ticket, i'll get this deleted

src/app/forms/custom-sections/plates/plates.component.ts
kpearmanbjss: Yup that's correct, originally it was used in a few places. We've just refactored them all out, i'll get that changed!

naathanbrown (approved):

Daniel-Searle (approved):

gjulien-bjss (approved):

PR #1180, opened 2023-09-25
feat(cb2-7045): create link on search result page

JunYanBJSS:
## feat(cb2-7045): create a new Tech Record when a search returns no results

Handle the case where no results return when doing a search
[CB2-7045](https://dvsa.atlassian.net/browse/CB2-7045)
## Checklist

### DIFF
--- a/src/app/features/search/multiple-search-results/multiple-search-results.component.html
+++ b/src/app/features/search/multiple-search-results/multiple-search-results.component.html
@@ -1,13 +1,19 @@
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-full">
-    <a class="govuk-back-link" id="multiple-search-results-back-link" (click)="navigateBack()">Back</a>
-  </div>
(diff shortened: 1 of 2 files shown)


gjulien-bjss (commented):
I would stick to the standard gov uk classes, unless there is a specific reason to change it. These classes and components have been studied to accessibility and are designed to feel standardised across government services.

You can take a look [here](https://design-system.service.gov.uk/components/back-link/) for example. 

Looks good otherwise

gjulien-bjss (commented):
You've also got some tests failing

JunYanBJSS:
> I would stick to the standard gov uk classes, unless there is a specific reason to change it. These classes and components have been studied to accessibility and are designed to feel standardised across government services.
> 
> You can take a look [here](https://design-system.service.gov.uk/components) for example. I also don't really think it should be a back link, it probably should just look like a normal link.
> 
> Looks good otherwise.

I got it!  I will look into these issues you mentioned in my PR. Thank you so much for your reviews.

PR #1186, opened 2023-09-28
fix(cb2-9195): fixes email validation for multiple-period domain addresses

andrewf-bjss:
Fixes the email validation regex to allow domain names with multiple periods (e.g. `.co.uk`)

[CB2-9195](https://dvsa.atlassian.net/browse/CB2-9195)

## Checklist

- [x] Delete branch after merge

### DIFF
--- a/src/app/forms/validators/custom-validators.ts
+++ b/src/app/forms/validators/custom-validators.ts
@@ -181,7 +181,7 @@ export class CustomValidators {
   }
 
   static email(): ValidatorFn {
-    return this.customPattern(['^[-\\w.\\+]+@[-\\w]+\\.[A-Za-z]{2,}$', 'Enter an email address in the correct format, like name@example.com']);
+    return this.customPattern(['^[\\w\\-\\.]+@([\\w-]+\\.)+[\\w-]{2,}$', 'Enter an email address in the correct format, like name@example.com']);
   }
 
   static customPattern([regEx, message]: string[]): ValidatorFn {
--- a/src/app/forms/validators/custom-validators.ts
+++ b/src/app/forms/validators/custom-validators.ts
@@ -181,7 +181,7 @@ export class CustomValidators {
   }
 
   static email(): ValidatorFn {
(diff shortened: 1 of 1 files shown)


kcarr-bjss (dismissed):

src/app/forms/validators/custom-validators.ts
```diff
@@ -181,7 +181,7 @@ export class CustomValidators {
   }
 
   static email(): ValidatorFn {
-    return this.customPattern(['^[-\\w.\\+]+@[-\\w]+\\.[A-Za-z]{2,}$', 'Enter an email address in the correct format, like name@example.com']);
+    return this.customPattern(['^[\\w\\-\\.]+@([\\w-]+\\.)+[\\w-]{2,}$', 'Enter an email address in the correct format, like name@example.com']);
```
gjulien-bjss, line 184: has the first `+` been removed for a reason? It stops [plus address email addresses](https://learn.microsoft.com/en-us/exchange/recipients-in-exchange-online/plus-addressing-in-exchange-online) such as `foo+bar@email.com` from being entered if you don't have it in there.

naathanbrown (approved):

src/app/forms/validators/custom-validators.ts
naathanbrown: Gunna miss this guy

andrewf-bjss: only because it wasn't in the validator I'd found - added back in and re-tested locally

gjulien-bjss (approved):
Assuming the tests etc... pass, all good!

src/app/forms/validators/custom-validators.ts
gjulien-bjss: I also found a email validator that didn't support them... I've found from trying to use them personally that about 25% of websites won't allow it... which is quite annoying.

shivangidas (approved):

andrewf-bjss:
@Twebb97 - yes, I think it does, but I think there's also a max length on our email address fields

PR #1234, opened 2023-10-27
feature(cb2-9604): Added Dimension warnings

JunYanBJSS:
## Dimension warnings

Added Dimension warning component for meeting the AC of ticket.
[CB2-9604](https://dvsa.atlassian.net/browse/CB2-9604)

## Checklist

### DIFF
--- a/src/app/forms/components/number-input/number-input.component.ts
+++ b/src/app/forms/components/number-input/number-input.component.ts
@@ -30,10 +30,10 @@ export class NumberInputComponent extends BaseControlComponent implements AfterC
   }
 
   shouldDisplayLengthWarning(): boolean {
-    return this.label === 'Length' && Number(this.value) > 12000;
+    return this.label === 'Length' && parseInt(this.value, 10) > 12000;
   }
   shouldDisplayWidthWarning(): boolean {
-    return this.label === 'Width' && Number(this.value) > 2600;
+    return this.label === 'Width' && parseInt(this.value, 10) > 2600;
   }
 
   isCorrectVehicleType(): boolean {
--- a/src/app/forms/components/number-input/number-input.component.ts
+++ b/src/app/forms/components/number-input/number-input.component.ts
@@ -1,4 +1,4 @@
-import { AfterContentInit, Component } from '@angular/core';
+import { AfterContentInit, Component, Input } from '@angular/core';
 import { NG_VALUE_ACCESSOR } from '@angular/forms';
 import { BaseControlComponent } from '../base-control/base-control.component';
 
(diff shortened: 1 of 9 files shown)


src/app/forms/components/number-input/number-input.component.ts
```diff
@@ -15,10 +15,31 @@ import { BaseControlComponent } from '../base-control/base-control.component';
...
+
+    if (this.isCorrectVehicleType()) {
+      if (this.shouldDisplayLengthWarning()) return 'This length dimension field value is greater than 12,000mm. Check your input before proceeding';
+      if (this.shouldDisplayWidthWarning()) return 'This width dimension field value is greater than 2,600mm. Check your input before proceeding';
+    }
+    return '';
+  }
+
+  shouldDisplayLengthWarning(): boolean {
+    return this.label === 'Length' && Number(this.value) > 12000;
```
pbardy2000, line 33: This is fine for your use case, but I would caution against using Number for processing what I assume are string values as it might not always work as expected, e.g. for the opposition condition:

Number('') < 12000  // true
Number(null) < 12000 // true

Whereas parseInt might be more inline with the behaviour you actually want, e.g:

parseInt(null) < 12000 // false
parseInt('') < 12000 // false

pbardy2000 (dismissed):

src/app/forms/components/number-input/number-input.component.ts
JunYanBJSS: That is a good point Bardy! I will take more attention in this similar situation.

shivangidas:
```suggestion
    return this.label === 'Length' && parseInt(this.value) > 12000;
```

src/app/forms/components/number-input/number-input.component.ts
```diff
@@ -15,10 +15,31 @@ import { BaseControlComponent } from '../base-control/base-control.component';
...
+  }
+  shouldDisplayWidthWarning(): boolean {
+    return this.label === 'Width' && Number(this.value) > 2600;
```
shivangidas, line 36:
```suggestion
    return this.label === 'Width' && parseInt(this.value) > 2600;
```

src/app/forms/components/field-warning-message/field-warning-message.component.scss
shivangidas, line 1: unnecessary file. can be removed

JunYanBJSS: Removed it now>+.

shivangidas (approved):

PR #1235, opened 2023-10-30
feat(cb2-9601): update vehicle configuration options

owen-corrigan-bjss:
CB2-9601

THE WORK FOR THIS TICKET IS NOT COMPLETE. This PR is to give everyone access to the resolver that null's tech record fields based on enums. 

The functionality for the ticket is complete but there are tests to write for the resolver and the enum needs swapping for the types package. 

_One line description_
[CB2-9601](https://dvsa.atlassian.net/browse/CB2-9601)

## Checklist

### DIFF
--- a/src/app/features/tech-record/components/tech-record-summary-changes/tech-record-summary-changes.component.ts
+++ b/src/app/features/tech-record/components/tech-record-summary-changes/tech-record-summary-changes.component.ts
@@ -205,7 +205,8 @@ export class TechRecordSummaryChangesComponent implements OnInit, OnDestroy {
   }
 
   isNotEmpty(value: unknown): boolean {
-    if (value == null || value === '') return false;
+    if (value === null) return true;
+    if (value === '') return false;
     if (typeof value === 'object') return Object.values(value).length > 0;
     return true;
   }
--- a/src/app/features/tech-record/components/tech-record-summary-changes/tech-record-summary-changes.component.ts
+++ b/src/app/features/tech-record/components/tech-record-summary-changes/tech-record-summary-changes.component.ts
@@ -205,7 +205,7 @@ export class TechRecordSummaryChangesComponent implements OnInit, OnDestroy {
   }
 
   isNotEmpty(value: unknown): boolean {
-    if (value === '') return false;
+    if (value === '' || value === undefined) return false;
     if (typeof value === 'object' && value !== null) return Object.values(value).length > 0;
     return true;
   }
--- a/src/app/features/tech-record/components/tech-record-summary-changes/tech-record-summary-changes.component.ts
+++ b/src/app/features/tech-record/components/tech-record-summary-changes/tech-record-summary-changes.component.ts
@@ -205,9 +205,8 @@ export class TechRecordSummaryChangesComponent implements OnInit, OnDestroy {
   }
 
   isNotEmpty(value: unknown): boolean {
-    if (value === null) return true;
     if (value === '') return false;
-    if (typeof value === 'object') return Object.values(value).length > 0;
+    if (typeof value === 'object' && value !== null) return Object.values(value).length > 0;
     return true;
   }
 
--- a/src/app/features/tech-record/tech-record-routing.module.ts
+++ b/src/app/features/tech-record/tech-record-routing.module.ts
@@ -5,6 +5,7 @@ import { CancelEditTechGuard } from '@guards/cancel-edit-tech/cancel-edit-tech.g
 import { RoleGuard } from '@guards/role-guard/roles.guard';
 import { Roles } from '@models/roles.enum';
 import { ReasonForEditing } from '@models/vehicle-tech-record.model';
+import { TechRecordReviewResolver } from 'src/app/resolvers/tech-record-review/tech-record-review.resolver';
 import { techRecordViewResolver } from 'src/app/resolvers/tech-record-view/tech-record-view.resolver';
 import { TechRecordAmendReasonComponent } from './components/tech-record-amend-reason/tech-record-amend-reason.component';
 import { AmendVinComponent } from './components/tech-record-amend-vin/tech-record-amend-vin.component';
@@ -18,9 +19,7 @@ import { GeneratePlateComponent } from './components/tech-record-generate-plate/
 import { TechRecordSearchTyresComponent } from './components/tech-record-search-tyres/tech-record-search-tyres.component';
 import { TechRecordSummaryChangesComponent } from './components/tech-record-summary-changes/tech-record-summary-changes.component';
 import { TechRecordUnarchiveComponent } from './components/tech-record-unarchive/tech-record-unarchive-component';
-import { TechRouterOutletComponent } from './components/tech-router-outlet/tech-router-outlet.component';
 import { TechRecordComponent } from './tech-record.component';
-import { TechRecordReviewResolver } from 'src/app/resolvers/tech-record-review/tech-record-review.resolver';
 
 const routes: Routes = [
   {
--- a/src/app/features/tech-record/tech-record-routing.module.ts
+++ b/src/app/features/tech-record/tech-record-routing.module.ts
@@ -5,7 +5,7 @@ import { CancelEditTechGuard } from '@guards/cancel-edit-tech/cancel-edit-tech.g
 import { RoleGuard } from '@guards/role-guard/roles.guard';
 import { Roles } from '@models/roles.enum';
 import { ReasonForEditing } from '@models/vehicle-tech-record.model';
-import { TechRecordReviewResolver } from 'src/app/resolvers/tech-record-review/tech-record-review.resolver';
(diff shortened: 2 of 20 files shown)


src/app/features/tech-record/components/tech-record-summary-changes/tech-record-summary-changes.component.ts
```diff
@@ -205,7 +205,8 @@ export class TechRecordSummaryChangesComponent implements OnInit, OnDestroy {
   }
 
   isNotEmpty(value: unknown): boolean {
-    if (value == null || value === '') return false;
+    if (value === null) return true;
```
naathanbrown, line 208: Would this not just fall into the return true at the bottom?

src/app/resolvers/tech-record-review/tech-record-review.resolver.ts
```diff
@@ -0,0 +1,112 @@
...
+import { selectTechRecord, updateEditingTechRecord } from '@store/technical-records';
+import {
+  catchError,
+  map,
+  of,
+  take,
+  tap,
+} from 'rxjs';
+
+export const TechRecordReviewResolver: ResolveFn<boolean> = () => {
```
naathanbrown, line 19: This is a function, so naming convention needs to be techRecordReviewResolver

src/app/features/tech-record/components/tech-record-summary-changes/tech-record-summary-changes.component.ts
```diff
@@ -205,8 +205,8 @@ export class TechRecordSummaryChangesComponent implements OnInit, OnDestroy {
   }
 
   isNotEmpty(value: unknown): boolean {
-    if (value == null || value === '') return false;
-    if (typeof value === 'object') return Object.values(value).length > 0;
+    if (value === '') return false;
+    if (typeof value === 'object' && value !== null) return Object.values(value).length > 0;
```
owen-corrigan-bjss, line 209: had to check that value is not null here otherwise it won't drop through to the final return

shivangidas (approved):

src/app/features/tech-record/tech-record-routing.module.ts
```diff
@@ -40,7 +40,10 @@ const routes: Routes = [
       isCustomLayout: true,
     },
     canActivate: [MsalGuard, RoleGuard],
-    resolve: { techRecord: techRecordViewResolver },
+    resolve: {
+      techRecord: techRecordViewResolver,
+      load: techRecordReviewResolver,
```
shivangidas, line 45: the resolver name could be changed maybe 😁

src/app/features/tech-record/components/tech-record-summary-changes/tech-record-summary-changes.component.ts
pbardy2000: Just want to double check if this is an issue or not, this checks for exact equality to null, and thus will allow undefined - is that the behaviour you want?

owen-corrigan-bjss: @pbardy2000 So we will only be wanting to let null through as the resolver is nulling the fields that contain bad data rather than setting them to undefined. I'll extend the empty string check to check for undefined

Twebb97: Is there a test for this that has an explicit null value ?

src/app/resolvers/tech-record-validate/tech-record-validate.resolver.ts
```diff
@@ -0,0 +1,112 @@
...
+    if (record[key as keyof TechRecordVehicleType<'hgv'>]) {
+      const valid = Object.values(key).includes(checks[`${key}`]);
+      if (!valid) {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        (validatedRecord as any)[`${key}`] = null;
+      }
+    }
+  });
+  return validatedRecord as TechRecordType<'put'>;
+};
```
pbardy2000, line 112:
```suggestion
const handlePsv = (record: TechRecordVehicleType<'psv'>) => {
  const validatedRecord: TechRecordVehicleType<'psv'> = { ...record };

  const checks = {
    techRecord_vehicleConfiguration: HgvPsvVehicleConfiguration,
  } as const;

  type Checks = typeof checks;
  type CheckKey = keyof Checks;

  const keys = Object.keys(checks) as CheckKey[];

  keys.forEach((key) => {
    if (record[key as keyof TechRecordVehicleType<'psv'>]) {
      const valid = Object.values(checks).includes(checks[key]);
      if (!valid) {
        validatedRecord[key] = null;
      }
    }
  });
  return validatedRecord as TechRecordType<'put'>;
};

const handleTrl = (record: TechRecordVehicleType<'trl'>) => {
  const validatedRecord: TechRecordVehicleType<'trl'> = { ...record };

  const checks = {
    techRecord_vehicleConfiguration: TrlVehicleConfiguration,
  };

  type Checks = typeof checks;
  type CheckKey = keyof Checks;

  const keys = Object.keys(checks) as CheckKey[];

  keys.forEach((key: CheckKey) => {
    if (record[key as keyof TechRecordVehicleType<'trl'>]) {
      const valid = Object.values(checks).includes(checks[key]);
      if (!valid) {
        validatedRecord[key] = null;
      }
    }
  });
  return validatedRecord as TechRecordType<'put'>;
};

const handleHgv = (record: TechRecordVehicleType<'hgv'>) => {
  const validatedRecord: TechRecordVehicleType<'hgv'> = { ...record };

  const checks = {
    techRecord_vehicleConfiguration: HgvPsvVehicleConfiguration,
  };

  type Checks = typeof checks;
  type CheckKey = keyof Checks;

  const keys = Object.keys(checks) as CheckKey[];

  keys.forEach((key: CheckKey) => {
    if (record[key as keyof TechRecordVehicleType<'hgv'>]) {
      const valid = Object.values(checks).includes(checks[key]);
      if (!valid) {
        validatedRecord[key] = null;
      }
    }
  });
  return validatedRecord as TechRecordType<'put'>;
};
```

Do you not want something more like this?

src/app/features/tech-record/components/tech-record-summary-changes/tech-record-summary-changes.component.ts
Twebb97: Ive got 2 things here:

1. lets add a test for an object with null to be allowed througj

2. We're now saying that we allow null but not empty in a function called isNotEmpty which would raise some questions in the future if I saw this again. In the object world this would break the O in solid, even more so that it was discussed that this would move into a more utils type file in the future.

Hence is there a more appropriate name and can we self document the code, or do we default to a comment ?

src/app/forms/custom-sections/body/body.component.ts
```diff
@@ -104,9 +105,18 @@ export class BodyComponent implements OnInit, OnChanges, OnDestroy {
   }
 
   get bodyTypes(): MultiOptions {
-    const optionsMap = vehicleBodyTypeCodeMap.get(this.techRecord.techRecord_vehicleType as VehicleTypes);
-    const values = [...optionsMap!.values()];
+    let vehicleType: string = this.techRecord.techRecord_vehicleType;
+
+    if (this.techRecord.techRecord_vehicleType === 'hgv' && this.techRecord.techRecord_vehicleConfiguration === 'articulated') {
```
Twebb97, line 110: Are we going to see more of this type of change in the future for other specific vehicle types ?
if so is there going to be a massive if else chain that can be avoided early ?

src/app/forms/validators/custom-validators.ts
```diff
@@ -298,4 +298,17 @@ export class CustomValidators {
       return null;
     };
   };
+
+  static updateFunctionCode = (): ValidatorFn => {
+    return (control: AbstractControl): ValidationErrors | null => {
+      const vehicleFunctionCode = control.root.get('techRecord_functionCode');
+      if (control.dirty) {
+        if (control?.value === 'rigid') vehicleFunctionCode?.setValue('R', { emitEvent: false });
```
Twebb97, line 306: You can avoid this using a map with types and thier function codes and someting like this 

vehicleFunctionCode?.setValue(functionCode.get(TYPE_VARIBALE), { emitEvent: false })

src/app/features/tech-record/components/tech-record-summary-changes/tech-record-summary-changes.component.ts
owen-corrigan-bjss: Both fair points Tom I'll work on those

shivangidas (approved):

src/app/forms/custom-sections/body/body.component.ts
shivangidas: looking at the ticket, it appears all vehicles have different conditions now for displaying the  https://dvsa.atlassian.net/browse/CB2-9601

PR #1238, opened 2023-10-30
feat(cb2-9183): update vehicle class

shivangidas:
## Update vehicle class options
Stricter validation for vehicle class

[CB2-9183](https://dvsa.atlassian.net/browse/CB2-9183)

## Checklist


src/app/resolvers/tech-record-validate/tech-record-validate.resolver.ts
```diff
@@ -76,6 +80,7 @@ const handlePsv = (record: TechRecordVehicleType<'psv'>) => {
 
 const handleTrl = (record: TechRecordVehicleType<'trl'>) => {
   const validatedRecord: TechRecordVehicleType<'trl'> = { ...record };
+  validatedRecord.techRecord_vehicleClass_description = VehicleClassDescription.Trailer;
```
shivangidas, line 83: Can't decide if I want to check if it is a trailer class before adding. AC is trailer must always be a trailer class

Twebb97: Probably fine as is, self documenting for trailers, the only considerations are if it would be inherently clear that this method would default the description and if anything would ever need it to not do that.

Twebb97 (approved):

kpearmanbjss (approved):

src/app/resolvers/tech-record-validate/tech-record-validate.resolver.ts
shivangidas: Good point, once all the validations are added this sprint, can come back to make this resolver more clear if needed

PR #1239, opened 2023-10-30
feat(cb2-9601): update vehicle configuration options

owen-corrigan-bjss:
CB2-9601

fixes issue with the resolver also adds unit tests to the resolver to check it's working correctly. Small change to the body component. 
 
_One line description_

## Checklist


shivangidas (approved):

src/app/forms/custom-sections/body/body.component.ts
```diff
@@ -153,4 +151,12 @@ export class BodyComponent implements OnInit, OnChanges, OnDestroy {
       this.optionsService.loadOptions(ReferenceDataResourceType.TrlMake);
     }
   }
+
+  updateHgvVehicleBodyType() {
+    if (this.techRecord.techRecord_vehicleType === 'hgv' && this.techRecord.techRecord_vehicleConfiguration === 'articulated') {
```
Twebb97, line 156: I think you already know the vehicle is hgv at this point from the parent calling, this is an interesting one. Again this looks like it will end up being a chain of conditional checks on vehicle configuration.

since really you're just passing it through arnt you actually just checking that its a valid configuration for a hgv and then you can handle anything thats not valid a specific way ?

shivangidas (approved):

src/app/resolvers/tech-record-validate/tech-record-validate.resolver.ts
```diff
@@ -8,20 +8,20 @@ import { HgvPsvVehicleConfiguration, TrlVehicleConfiguration } from '@models/veh
 import { Store } from '@ngrx/store';
 import { State } from '@store/.';
 import { selectTechRecord, updateEditingTechRecord } from '@store/technical-records';
+import * as _ from 'lodash';
```
naathanbrown, line 11: I don't think we should import this way from lodash if possible not to. Import via function

shivangidas (approved):

Twebb97 (approved):

naathanbrown (approved):

PR #1246, opened 2023-11-02
feat(cb2-9601): update vehicle configuration

owen-corrigan-bjss:
CB2-9601: 

fixes an issue with the resolver not working on a refresh, navigates you back to the original record. 

_One line description_

## Checklist

### DIFF
--- a/src/app/resolvers/tech-record-validate/tech-record-validate.resolver.ts
+++ b/src/app/resolvers/tech-record-validate/tech-record-validate.resolver.ts
@@ -23,9 +23,7 @@ import { State } from '@store/.';
 import { selectTechRecord, updateEditingTechRecord } from '@store/technical-records';
 import { isEqual } from 'lodash';
 import {
-  catchError,
   map,
-  of,
   take,
 } from 'rxjs';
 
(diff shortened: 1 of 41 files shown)


src/app/resolvers/tech-record-validate/tech-record-validate.resolver.ts
```diff
@@ -59,9 +68,6 @@ export const techRecordValidateResolver: ResolveFn<boolean> = () => {
     map(() => {
       return true;
     }),
-    catchError(() => {
```
Twebb97, line 62: Can you check the rsjx docs for the errors store.select could throw as assumingly if the store breaks we would still need this ?

src/app/resolvers/tech-record-view/tech-record-view.resolver.spec.ts
```diff
@@ -85,20 +85,5 @@ describe('TechRecordViewResolver', () => {
 
       expect(dispatchSpy).toHaveBeenCalledTimes(2);
     });
-
-    it('should resolved to false if \'fetchTestResultsBySystemNumberFailed\' and \'getTechRecordV3Failure\' action are dipatched', () => {
```
Twebb97, line 89: what should it do now then ?

src/app/resolvers/tech-record-validate/tech-record-validate.resolver.ts
owen-corrigan-bjss: So looking at the docs select doesn't throw an error, also we're not throwing an error ourselves. selectTechRecord is always returning either techRecord or editingTechRecord and techRecord is initially undefined. There's now a check for if there is no tech record we navigate back

pbardy2000 (approved):

kpearmanbjss (approved):

PR #1257, opened 2023-11-10
feat(cb2-9728): add applicant details to adr section on a technical record

owen-corrigan-bjss:
CB2-9728:

First ticket for the ADR section. 

Adds the section for HGV, LGV and TRL. 
Will conditionally render the address section based on answer to initial radio button.   
warns users if they're going to delete the existing data.


kpearmanbjss (approved):

angular.json
```diff
@@ -12,8 +12,7 @@
         "@schematics/angular:application": {
           "strict": true
         },
-        "@schematics/angular:resolver": {
-          "functional": true,
+        "@schematics/angular:directive": {
```
naathanbrown, line 15: Why has this changed? Just interested to know

src/app/features/tech-record/components/tech-record-summary-changes/tech-record-summary-changes.component.ts
```diff
@@ -150,6 +151,7 @@ export class TechRecordSummaryChangesComponent implements OnInit, OnDestroy {
       .pipe(take(1), takeUntil(this.destroy$))
       .subscribe(([systemNumber, createdTimestamp]) => {
         if (systemNumber && createdTimestamp) {
+          this.store$.dispatch(clearADRDetailsBeforeUpdate());
```
naathanbrown, line 154: why are we clearing here?

src/app/features/tech-record/components/tech-record-summary-changes/tech-record-summary-changes.component.ts
```diff
@@ -205,8 +207,8 @@ export class TechRecordSummaryChangesComponent implements OnInit, OnDestroy {
   }
 
   isNotEmpty(value: unknown): boolean {
-    if (value === '' || value === undefined) return false;
-    if (typeof value === 'object' && value !== null) return Object.values(value).length > 0;
+    if (value === '' || value === undefined || value === null) return false;
```
naathanbrown, line 210: if(value) should do the same as this

src/app/forms/components/dynamic-form-field/dynamic-form-field.component.ts
```diff
@@ -3,7 +3,7 @@ import { AfterContentInit, Component, Input } from '@angular/core';
 import { FormGroup } from '@angular/forms';
 import { CustomFormControl, FormNodeEditTypes, FormNodeOption } from '@forms/services/dynamic-form.types';
 import { MultiOptionsService } from '@forms/services/multi-options.service';
-import { Observable, of, map } from 'rxjs';
+import { Observable, map, of } from 'rxjs';
```
naathanbrown, line 6: great change (love the linter haha)

src/app/forms/custom-sections/adr/adr.component.ts
```diff
@@ -0,0 +1,60 @@
...
+})
+export class AdrComponent implements OnInit {
+  @Input() techRecord!: TechRecordType<'hgv'> | TechRecordType<'trl'> | TechRecordType<'lgv'>;
+  @Input() isEditing = false;
+  @Input() disableLoadOptions = false;
+  @Output() formChange = new EventEmitter();
+
+  public template = AdrTemplate;
+  public form!: CustomFormGroup;
+  public adrDetails = [
```
naathanbrown, line 22: I worry about this, we now basically have some magic strings to check in a component, can this maybe be driven by the templates like the form is? The init creates the form, so you have the controls created etc

naathanbrown: Hmm think it would maybe need to be the template not the form

src/app/forms/directives/nesting-level/nesting-level.directive.ts
```diff
@@ -0,0 +1,29 @@
...
+  selector: '[appNestingLevel]',
+})
+export class NestingLevelDirective implements OnInit {
+  @Input() nestingLevel = 0;
+
+  constructor(public readonly elementRef: ElementRef<HTMLElement>) {
+  }
+
+  ngOnInit(): void {
+    const ml = this.nestingLevel > 1 ? (2 * (this.nestingLevel - 1)) : 0;
```
naathanbrown, line 15: what are these consts?

src/app/forms/components/dynamic-form-field/dynamic-form-field.component.html
```diff
@@ -1,7 +1,9 @@
-<ng-container
+<div
   *ngIf="form && control && control.value.meta.editType !== formNodeEditTypes.HIDDEN"
   [ngSwitch]="control.value.meta.editType"
   [formGroup]="form"
+  appNestingLevel
+  [nestingLevel]="$any(form).meta.nestingLevel ?? 0"
```
BrandonT95, line 6: You can simultaneously apply the directive and nesting level at once if you rename input property to appNestingLevel within the directive

angular.json
pbardy2000: Think the resolver bit must have got deleted when resolving merge conflicts, so I've added that back - changed the directive to not be flat so it, and its spec files sit in the same folder, just like components and services.

src/app/features/tech-record/components/tech-record-summary-changes/tech-record-summary-changes.component.ts
pbardy2000: True, have just realised there is a isEmpty function in lodash which does exactly this, so will substitute that in instead.

src/app/forms/components/dynamic-form-field/dynamic-form-field.component.html
pbardy2000: Didn't know that, thanks. Will modify.

src/app/forms/custom-sections/adr/adr.component.ts
pbardy2000: Yeah, this is a symptom of the way we're handling the migrated data - the old adr data doesn't have the dangerous goods field, and all the other properties lie flat on the object. We need the field as this determines whether the dropdown is open, and when amending, if other adr fields should be nulled. I raised this with approval types on a previous ticket, because it seems weird to be that we'd do this kind of data correction on the front-end and not during the migration of data, especially it doesn't require any user input.

Not sure how this kind of logic could be implemented solely by the templates though...

src/app/forms/directives/nesting-level/nesting-level.directive.ts
pbardy2000: lol. Sorry, I've been using tailwind for so long these looked normal: margin left, padding left, margin bottom, padding top, and border width.  Essentially this adds a grey border and indentation proportional to the nesting level, which is hard to do by simply adding a css class, so this modifies css properties directly though js. This directive might be a bit of gold plating, and not actually needed. The designs we were given are subject to change, and they might not use the indentation thing in the final version, we just wanted to ensure that our approach was capable of handling the same level of nesting/complexity as previous versions.

naathanbrown (dismissed):

tomevs88 (approved):

PR #1288, opened 2023-11-28
fix(cb2-9937): review screen is broken

owen-corrigan-bjss:
CB2-9937:

The addition of the lodash isEmpty function broke the summary of changes page. This PR reverts that change and changes the orders of the resolvers on the route to summary changes, this was because on a refresh the page was loading as a blank page, rather than navigating back as it used to. 

## Ticket title

_One line description_

## Checklist

### DIFF
--- a/src/app/features/tech-record/components/tech-record-summary-changes/tech-record-summary-changes.component.ts
+++ b/src/app/features/tech-record/components/tech-record-summary-changes/tech-record-summary-changes.component.ts
@@ -28,7 +28,6 @@ import {
   updateTechRecord,
   updateTechRecordSuccess,
 } from '@store/technical-records';
-import { isEmpty } from 'lodash';
 import {
   Subject, combineLatest, map, take, takeUntil,
 } from 'rxjs';
--- a/src/app/features/tech-record/components/tech-record-summary-changes/tech-record-summary-changes.component.ts
+++ b/src/app/features/tech-record/components/tech-record-summary-changes/tech-record-summary-changes.component.ts
@@ -88,6 +88,7 @@ export class TechRecordSummaryChangesComponent implements OnInit, OnDestroy {
       .select(editingTechRecord)
       .pipe(take(1), takeUntil(this.destroy$))
       .subscribe((data) => {
+        if (!data) this.cancel();
         this.techRecordEdited = data;
       });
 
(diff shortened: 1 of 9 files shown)


src/app/features/tech-record/components/tech-record-summary-changes/tech-record-summary-changes.component.ts
```diff
@@ -96,6 +96,10 @@ export class TechRecordSummaryChangesComponent implements OnInit, OnDestroy {
       .pipe(take(1), takeUntil(this.destroy$))
       .subscribe((changes) => {
         this.techRecordChanges = changes;
+        if (this.vehicleType === VehicleTypes.PSV || this.vehicleType === VehicleTypes.HGV) {
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          delete (this.techRecordChanges as any).techRecord_numberOfWheelsDriven;
+        }
```
owen-corrigan-bjss, line 102: I know this looks weird but numberOfWheelsDriven is in the template for HGV and PSV (even though they're hidden) it ends up appearing as an empty field in vehicle summary. it seems to be something to do with tests and I don't want to break them for the sake of this bug.

src/app/features/tech-record/components/tech-record-summary-changes/tech-record-summary-changes.component.ts
```diff
@@ -96,6 +96,10 @@ export class TechRecordSummaryChangesComponent implements OnInit, OnDestroy {
       .pipe(take(1), takeUntil(this.destroy$))
       .subscribe((changes) => {
         this.techRecordChanges = changes;
+        if (this.vehicleType === VehicleTypes.PSV || this.vehicleType === VehicleTypes.HGV) {
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          delete (this.techRecordChanges as any).techRecord_numberOfWheelsDriven;
```
pbardy2000, line 101:
```suggestion
delete (this.techRecordChanges as Partial<TechRecordGETPSV | TechRecordGETHGV>).techRecord_numberOfWheelsDriven;
```

Yeah the getter for the vehicle type makes it impossible for that if statement to narrow the vehicle get type as you're not directly evaluating the techRecordChanges object, but I think this would be better as it doesn't introduce more eslint warnings.

owen-corrigan-bjss: Ah nice. I didn't realise that would work with it not being on the hgv or psv type. Gave it a try though and it seems good. Good to know

shivangidas (approved):

PR #1311, opened 2023-12-08
Feat(cb2-9745):  Add Declarations seen to ADR section on a Technical Record

JunYanBJSS:
## Add Declarations seen to ADR section on a Technical Record

I want to add the Declarations seen in VTM for an ADR on HGVs, LGVs, and TRLs

So that information about Declarations seen for an ADR can be captured, without using multiple legacy IT systems.
[CB2-9745](https://dvsa.atlassian.net/browse/CB2-9745)

## Checklist


src/app/forms/templates/general/adr.template.ts
```diff
@@ -326,5 +326,102 @@ export const AdrTemplate: FormNode = {
...
+      groups: ['weight_section', 'dangerous_goods'],
+      hide: true,
+      validators: [
+        { name: ValidatorNames.MaxLength, args: 8 },
+        { name: ValidatorNames.Numeric },
+        {
+          name: ValidatorNames.RequiredIfNotHidden,
+          args: {
+            sibling: 'techRecord_adrDetails_brakeEndurance',
+          },
```
tomcrawleyy, line 412: not sure how i feel about the naming of this, when it comes to the implementation it looks a little strange, the validator title gives off the impression that it would not take any arguments to me, i feel like it needs to reference the sibling in the name for example "RequiredIfSiblingTrue" or "RequiredIfSiblingEquals" which takes in a sibling and a value to compare against

tomcrawleyy, line 412: we should find a way to avoid doing this, it seems really strange that you would need to pass a sibling into a validator which should only care about whether its hidden or not

src/app/forms/custom-sections/adr/adr.component.ts
```diff
@@ -69,6 +74,7 @@ export class AdrComponent implements OnInit {
   }
 
   handleFormChange(event: Record<string, unknown>) {
+    console.log(event);
```
tomcrawleyy, line 77: remove this console log and then happy to approve

tomcrawleyy (approved):

PR #1319, opened 2023-12-13
feat(cb2-9782): feature flag ADR for dev, int, preprod, prod

shivangidas:
## Feature toggle for ADR section on a Technical Record

[CB2-9782](https://dvsa.atlassian.net/browse/CB2-9782)

## Checklist


.env.example
```diff
@@ -1,4 +1,7 @@
 production=false
+isDevelop=true
+isIntegration=false
+isPreProd=false
```
tomcrawleyy, line 4: just a thought - could we have a single variable to encapsulate this, something like environment: where the value is set to 'develop' 'integration' or 'preprod' instead of 4 separate properties?

shivangidas: ah my bad! didn't have to add this, those values are calculated based on the environment in setenv.ts

pbardy2000 (dismissed):

src/environments/environment.ts
```diff
@@ -5,6 +5,8 @@
 export const environment = {
   production: false,
   isDevelop: false,
+  isIntegration: false,
+  isPreProd: false,
```
tomcrawleyy, line 9: this ought to be one variable named environment: which contains the properties 'production', 'develop', 'integration' and 'preprod'

shivangidas: Agreed, changed to use environment

pbardy2000 (approved):

JunYanBJSS (approved):

tomcrawleyy (approved):

PR #1340, opened 2023-12-19
Hard coding branch for higher non-prod environments

hamiltonrossbjss:
## Ticket title

Hard coding branch for higher non-prod environments

## Checklist

### DIFF
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -13,4 +13,5 @@ jobs:
       bucket: '.prod'
     secrets:
       AWS_MGMT_ROLE: ${{ secrets.AWS_MGMT_ROLE }}
+      AWS_ADMIN_ROLE: ${{ secrets.AWS_PROD_ROLE }}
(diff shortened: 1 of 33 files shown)


thisisthetechie (approved):

.github/workflows/deploy-prod.yml
```diff
@@ -0,0 +1,17 @@
...
+run-name: "${{ github.actor }} - ${{ github.ref_name }} - ${{ github.run_id }} 🚀"
+
+on:
+  workflow_dispatch:
+
+jobs:
+  vtm-app:
+    uses: ./.github/workflows/deploy.yml
+    with:
+      branch: 'master'
```
JoshCarter-ops, line 11: what if we wanted to deploy from a release branch?

hamiltonrossbjss: Current Jenkins VTM deploys are always from master. Would make sense to be releasing from a release branch, but i think we want to have a wider discussion about release process, as that break the current pattern.

JoshCarter-ops (approved):

PR #1341, opened 2023-12-20
fix(cb2-10240): multiple warnings for same fields

shivangidas:
## Error message displayed twice in Batch creation

[CB2-10240](https://dvsa.atlassian.net/browse/CB2-10240)

## Checklist

### DIFF
--- a/src/app/features/tech-record/components/tech-record-summary/tech-record-summary.component.ts
+++ b/src/app/features/tech-record/components/tech-record-summary/tech-record-summary.component.ts
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-non-null-assertion */
 import { ViewportScroller } from '@angular/common';
 import {
   ChangeDetectionStrategy,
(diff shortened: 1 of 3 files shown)


pbardy2000 (approved):

src/app/forms/components/text-input/text-input.component.html
```diff
@@ -6,7 +6,7 @@
 
   <div *ngIf="hint" id="{{ name }}-hint" class="govuk-hint">{{ hint }}</div>
 
-  <app-field-error-message [error]="error ? error : warning" [name]="name"></app-field-error-message>
+  <app-field-error-message [error]="error" [name]="name"></app-field-error-message>
   <app-field-error-message [error]="warning" [name]="name"></app-field-error-message>
```
tomcrawleyy, line 10: would it make more sense to keep the top one and remove the bottom two? or does the ternary operator not evaluate how we'd expect?

shivangidas: Either works, but I made it match all the other input components

tomcrawleyy: If you're matching the other input components, the last one needs to be an app-field-warning-message, not error

pbardy2000 (approved):

src/app/features/tech-record/components/tech-record-summary/tech-record-summary.component.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
tomcrawleyy, line 1: this can be removed, the psvBrakes and trlBrakes wont cause a non null assertion if you amend lines 60 and 61 to use the ! operator instead of ?, then all instances of this.trlBrakes! and this.psvBrakes! can have the ! removed from lines 206, 211 and 212

tomcrawleyy (approved):

pbardy2000 (approved):

PR #1346, opened 2024-01-03
fix(cb2-10321): ensure validation is run only when control is visible

pbardy2000:
## Small fix

_One line description_
[CB2-10321](https://dvsa.atlassian.net/browse/CB2-10321)

## Checklist


owen-corrigan (approved):

shivangidas (approved):

src/app/forms/validators/custom-validators.ts
kpearmanbjss, line 1: Does this not mean that you could write an invalid date then just hide the control to pass the validator?

PR #1355, opened 2024-01-12
fix(cb2-10441): fix reference number incorrectly display

JunYanBJSS:
## ADR-Reference number display incorrectly
The reference number should be displayed when the user selects option 'yes' in the battery applicable field.

[CB2-10441](https://dvsa.atlassian.net/browse/CB2-10441)

## Checklist

### DIFF
--- a/src/app/forms/templates/general/adr.template.ts
+++ b/src/app/forms/templates/general/adr.template.ts
@@ -649,7 +649,7 @@ export const AdrTemplate: FormNode = {
       value: null,
       type: FormNodeTypes.CONTROL,
       width: FormNodeWidth.L,
-      groups: ['battery_list_applicable', 'battery_list_hide', 'dangerous_goods'],
+      groups: ['battery_list_applicable', 'battery_list_hide'],
       validators: [
         { name: ValidatorNames.MaxLength, args: 8 },
         {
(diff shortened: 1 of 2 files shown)


tomevs88 (approved):

src/app/forms/templates/general/adr.template.ts
```diff
@@ -649,7 +649,7 @@ export const AdrTemplate: FormNode = {
       value: null,
       type: FormNodeTypes.CONTROL,
       width: FormNodeWidth.L,
-      groups: ['battery_list_applicable', 'battery_list_hide', 'dangerous_goods'],
+      groups: ['battery_list_applicable', 'battery_list_hide'],
```
pbardy2000, line 652: Not sure if its this simple, if you click yes on battery list applicable, then click 'no' on dangerous goods then the reference number will still be shown when its meant to be hidden.

JunYanBJSS: Ahhhh, you are right. Nice shout out. I didn't realize the test scenario you mentioned before.

pbardy2000 (dismissed):

pbardy2000 (approved):

tomcrawleyy (approved):

PR #1357, opened 2024-01-15
fix(cb2-10112): make max length validator actually work

pbardy2000:
## Ticket title

_One line description_
[CB2-10112](https://dvsa.atlassian.net/browse/CB2-10112)

## Checklist


JunYanBJSS (approved):

src/app/forms/custom-sections/adr-tank-statement-un-number-edit/adr-tank-statement-un-number-edit.component.ts
```diff
@@ -92,25 +92,31 @@ export class AdrTankStatementUnNumberEditComponent extends CustomFormControlComp
     this.formArray.controls.forEach((control, index) => {
       // Make all UN NUmbers labels reflect their position in form array
       control.meta.customId = `${this.control?.meta.name}_${index + 1}`;
-      control.meta.customErrorMessage = `UN number ${index + 1} is required or remove UN number ${index + 1}`;
+      control.meta.customErrorMessage = control.invalid && control.hasError('maxlength')
+        ? `UN number ${index + 1} must be less than or equal to 1500 characters`
+        : `UN number ${index + 1} is required or remove UN number ${index + 1}`;
```
tomcrawleyy, line 97: could this cause the required error to show when the control is valid?

pbardy2000: No, the custom error message is a meta property which determines the error label text, but will only be shown if the control is invalid. TBH, this code is a bit hacky but was the only way I could get the error messages of the child elements of custom form arrays to show up properly in the global error service, before hand when it would show the UN Number 1 is required or remove UN Number 1 error message when the length exceeded 1500, because it was using the custom error message property. This also won't overwrite the other one, because they're mutually exclusive.

tomcrawleyy (approved):

PR #1380, opened 2024-02-08
feat(cb2-9990): update feature flags for adr epics

shivangidas:
## Show ADR data entry in VTM in pre-prod, Hide ADR in VTA, Hide ADR certificate generation in VTM

[CB2-9990](https://dvsa.atlassian.net/browse/CB2-9990)

## Checklist

### DIFF
--- a/src/assets/featureToggle.prod.json
+++ b/src/assets/featureToggle.prod.json
@@ -1,4 +1,5 @@
 {
   "testToggle": false,
-  "adrToggle": false
+  "adrToggle": true,
+  "adrCertToggle": false
 }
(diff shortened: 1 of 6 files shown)


src/assets/featureToggle.prod.json
```diff
@@ -1,4 +1,5 @@
 {
   "testToggle": false,
-  "adrToggle": false
+  "adrToggle": true,
```
tomcrawleyy, line 3: do we need to keep the adrToggle flag in this json once its all enabled? since the flags have been removed from the app i would have expected this to follow suit

shivangidas: the flags haven't been removed, there was some extra floating around unused in the vehicle tech record component which I removed. I agree once it is ON in Prod we need to retire the flags completely. will add something in the backlog to do that

tomcrawleyy (approved):

pbardy2000 (approved):

PR #1385, opened 2024-02-15
feat(cb2-10930): refactor un subsequent inspections component

pbardy2000:
## ADR Product List: UN Number 1 validation broken, allowing submission of UN Numbers over max length

- Refactors the tank statement un numbers edit component - moving validation logic back to the template.
- Removes the old validator in favour of an inline custom one for both UN numbers and the reference number
- Adds functionality to the global error service to display multiple separate errors from a single validator - this was needs to show multiple warnings of controls of form array exceeding the 1500 character limit.

[CB2-10930](https://dvsa.atlassian.net/browse/CB2-10930)

## Checklist

### DIFF
--- a/src/app/forms/custom-sections/adr-tank-statement-un-number-edit/adr-tank-statement-un-number-edit.component.html
+++ b/src/app/forms/custom-sections/adr-tank-statement-un-number-edit/adr-tank-statement-un-number-edit.component.html
@@ -32,7 +32,7 @@
               ></app-text-input>
             </td>
             <td class="govuk-table__cell align-middle">
-              <a class="govuk-link" role="button" href="javascript:void(0)" (click)="removeControl(i)">Remove</a>
+              <a class="govuk-link link" (click)="removeControl(i)">Remove</a>
             </td>
           </tr>
         </ng-container>
@@ -41,6 +41,6 @@
   </ng-template>
 
   <p class="govuk-body">
-    <a role="button" class="govuk-link" href="javascript:void(0)" (click)="addControl()"> Add UN Number </a>
+    <a class="govuk-link link" (click)="addControl()">Add UN Number</a>
   </p>
 </ng-container>
--- a/src/app/forms/custom-sections/adr-tank-statement-un-number-edit/adr-tank-statement-un-number-edit.component.html
+++ b/src/app/forms/custom-sections/adr-tank-statement-un-number-edit/adr-tank-statement-un-number-edit.component.html
@@ -1,40 +1,44 @@
-<ng-container *ngIf="control && formArray.controls[0]">
-  <app-field-error-message
-    [error]="!formArray.controls[0].value && control.invalid && (control.touched || submitted) ? control.meta.customErrorMessage : ''"
-  ></app-field-error-message>
+<ng-container *ngIf="control && formArray.controls.length">
+  <app-field-error-message [error]="control.touched ? control.getError('custom')?.message : ''"></app-field-error-message>
 
-  <ng-container *ngIf="formArray.controls.length === 1">
+  <ng-container *ngIf="formArray.controls.length === 1; else multiple">
     <app-text-input
-      [id]="control.meta.name"
-      [name]="formArray.controls[0].meta.name"
-      [label]="formArray.controls[0].meta.label + ' 1'"
-      [customId]="formArray.controls[0].meta.customId"
+      [id]="'UN_number_1'"
+      [name]="'UN number 1'"
+      [label]="'UN number 1'"
+      [customId]="'UN number 1'"
       [formControl]="formArray.controls[0]"
+      [minlength]="1"
+      [maxlength]="1500"
       (blur)="formArray.controls[0].markAsTouched(); control.markAsTouched()"
     ></app-text-input>
   </ng-container>
 
-  <!-- Additional UN numbers -->
-  <table class="govuk-table" *ngIf="formArray.controls.length > 1">
-    <tbody class="govuk-table__body">
-      <ng-container *ngFor="let child of formArray.controls; let i = index">
-        <tr>
-          <td class="govuk-table__cell">
-            <app-text-input
-              [name]="child.meta.name"
-              [label]="(child.meta.label ?? '') + ' ' + (i + 1)"
-              [customId]="child.meta.customId"
-              [formControl]="child"
-              (blur)="child.markAsTouched(); control.markAsTouched()"
-            ></app-text-input>
-          </td>
-          <td class="govuk-table__cell align-middle">
-            <a class="govuk-link" role="button" href="javascript:void(0)" (click)="removeControl(i)">Remove</a>
-          </td>
-        </tr>
-      </ng-container>
-    </tbody>
-  </table>
+  <ng-template #multiple>
+    <table class="govuk-table">
+      <tbody class="govuk-table__body">
+        <ng-container *ngFor="let child of formArray.controls; let i = index">
+          <tr>
+            <td class="govuk-table__cell">
+              <app-text-input
+                [id]="'UN_number_' + (i + 1)"
+                [name]="'UN number ' + (i + 1)"
+                [label]="'UN number ' + (i + 1)"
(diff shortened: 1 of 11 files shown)


src/app/forms/custom-sections/adr-tank-statement-un-number-edit/adr-tank-statement-un-number-edit.component.html
```diff
@@ -1,40 +1,44 @@
...
-    </tbody>
-  </table>
+  <ng-template #multiple>
+    <table class="govuk-table">
+      <tbody class="govuk-table__body">
+        <ng-container *ngFor="let child of formArray.controls; let i = index">
+          <tr>
+            <td class="govuk-table__cell">
+              <app-text-input
+                [id]="'UN_number_' + (i + 1)"
```
owen-corrigan-bjss, line 24: Will this not duplicate the first ID? seen as you have UN_number_1 set on line 6?

src/app/forms/custom-sections/adr-tank-statement-un-number-edit/adr-tank-statement-un-number-edit.component.html
```diff
@@ -1,40 +1,44 @@
...
+                [label]="'UN number ' + (i + 1)"
+                [customId]="'UN number ' + (i + 1)"
+                [formControl]="child"
+                [minlength]="1"
+                [maxlength]="1500"
+                (blur)="child.markAsTouched(); control.markAsTouched()"
+              ></app-text-input>
+            </td>
+            <td class="govuk-table__cell align-middle">
+              <a class="govuk-link" role="button" href="javascript:void(0)" (click)="removeControl(i)">Remove</a>
```
naathanbrown, line 35: No to javascript void please, recommend looking up why it's bad and using an alternative.

src/app/forms/custom-sections/adr-tank-statement-un-number-edit/adr-tank-statement-un-number-edit.component.html
```diff
@@ -1,40 +1,44 @@
...
+  <app-field-error-message [error]="control.touched ? control.getError('custom')?.message : ''"></app-field-error-message>
 
-  <ng-container *ngIf="formArray.controls.length === 1">
+  <ng-container *ngIf="formArray.controls.length === 1; else multiple">
     <app-text-input
-      [id]="control.meta.name"
-      [name]="formArray.controls[0].meta.name"
-      [label]="formArray.controls[0].meta.label + ' 1'"
-      [customId]="formArray.controls[0].meta.customId"
+      [id]="'UN_number_1'"
```
naathanbrown, line 6: pure curiosity, why didn't this work using the meta stuff?

src/app/forms/custom-sections/adr-tank-statement-un-number-edit/adr-tank-statement-un-number-edit.component.html
owen-corrigan-bjss: obviously the same goes for the other fields

src/app/forms/custom-sections/adr-tank-statement-un-number-edit/adr-tank-statement-un-number-edit.component.html
naathanbrown: oh I see, ignore

pbardy2000: Ah, long story, this is a form array where the controls can be added or removed, theoretically out of order. The meta property is passed when constructing the custom form control, so when you add/remove a control you'd need to manually update all the controls labels so it has the correct number (to make clicking on global errors go to the correct text input). Doing it in html make it so that this happens automatically - e.g. adding a control causes a re-render so all the input bindings update properly.

src/app/forms/custom-sections/adr-tank-statement-un-number-edit/adr-tank-statement-un-number-edit.component.html
naathanbrown: isn't it an if else? so the top one won't render

pbardy2000: This is part of an if-else statement (done in Angular's control flow using *ngIf="con; else ngTemplate" - essentially we show a standalone UN number 1 when there is 1 control in the form array because this one should have a remove button and thus is full width - and no it a table. If we have more than 1 we render them all in a table, and all controls, including UN number 1 has a remove link.

owen-corrigan-bjss: Ah right sorry missed that

src/app/forms/custom-sections/adr-tank-statement-un-number-edit/adr-tank-statement-un-number-edit.component.html
pbardy2000: I disagree with the online reasons for why this is bad - they seem to be complaining about how IE handles this and alternatives like href="#" is even worse because it scrolls to the top of the page, but I agree that semantically this is not ideal. This was kind of a hack because the govuk-link style doesn't apply properly for anchor tags when no href is given, but will change to via css as that's probably better.

naathanbrown: in angular can you not just {{ return false }} or something?

naathanbrown: I also agree href=# is bad haha

pbardy2000: in angular can you not just {{ return false }} or something? - no clue - changed it to use css as that was what other people did in other places.

src/app/forms/custom-sections/adr-tank-statement-un-number-edit/adr-tank-statement-un-number-edit.component.spec.ts
```diff
@@ -112,8 +82,11 @@ describe('AdrTankStatementUnNumberEditComponent', () => {
     });
 
     it('should prevent the adding of additional controls, when the previous one is empty', () => {
+      component.formArray = fb.array<CustomFormControl>([]);
+      component.addControl('valid');
       component.addControl();
-      expect(component.formArray.value).toHaveLength(1);
+      component.addControl();
+      expect(component.formArray.value).toHaveLength(2);
     });
```
tomcrawleyy, line 90: i think this test description needs updating to encapsulate the behaviour of this test now as its not checking to see if adding an extra control is prevented, but rather expecting the opposite and that it has been added

src/app/forms/templates/general/adr.template.ts
```diff
@@ -491,13 +500,56 @@ export const AdrTemplate: FormNode = {
...
       groups: ['productList', 'statement_select_hide', 'tank_details_hide', 'dangerous_goods'],
       hide: true,
-      customErrorMessage: 'Reference number or UN number 1 is required when selecting Product List',
       validators: [
         {
-          name: ValidatorNames.RequiredIfEquals,
-          args: { sibling: 'techRecord_adrDetails_tank_tankDetails_tankStatement_productListRefNo', value: [null, undefined, ''] },
+          name: ValidatorNames.Custom,
+          args: (control: CustomFormControl) => {
+            if (control.meta?.hide) {
```
tomcrawleyy, line 507: is there a better place for this logic to sit instead of inside of the template? feels like this breaks the nice readable flow that is in place and if we start adding similar solutions to this it could get quite cluttered

tomcrawleyy (approved):

owen-corrigan-bjss (approved):

PR #1404, opened 2024-02-26
feat(cb2-9659): delete empty required standards object from payload

naathanbrown:
## Ticket title

Remove the empty array if we are an IVA pass.
[CB2-9659](https://dvsa.atlassian.net/browse/CB2-9659)

## Checklist

### DIFF
--- a/src/app/store/test-records/selectors/test-records.selectors.ts
+++ b/src/app/store/test-records/selectors/test-records.selectors.ts
@@ -48,15 +48,10 @@ export const testResultInEdit = createSelector(testResultsFeatureState, (state) 
 export const cleanedTestResultInEdit = createSelector(testResultsFeatureState, (state) => {
   if (state.editingTestResult?.testTypes?.at(0)) {
     const { testTypeId, requiredStandards } = state.editingTestResult.testTypes[0];
-    console.log((TEST_TYPES_GROUP1_SPEC_TEST.includes(testTypeId)));
-    console.log((TEST_TYPES_GROUP5_SPEC_TEST.includes(testTypeId)));
-    console.log(!(requiredStandards ?? []).length);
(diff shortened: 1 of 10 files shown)


src/app/store/test-records/selectors/test-records.selectors.ts
```diff
@@ -44,6 +45,16 @@ export const selectedTestResultState = createSelector(
 
 export const testResultInEdit = createSelector(testResultsFeatureState, (state) => state.editingTestResult);
 
+export const cleanedTestResultInEdit = createSelector(testResultsFeatureState, (state) => {
```
tomcrawleyy, line 48: not sure this is the best way to approach this, as this is going to scatter logic which is removing data from the tech record before submission, think it makes more sense to follow the same path that we took for ADR stuff, which is to dispatch an action which calls a reducer which updates the tech record right before submission

tomcrawleyy: for reference, this action which is ADR specific, is dispatched on line 170 in the tech-record-summary-changes component

naathanbrown: I thought about this, but the only way to make this work is to refactor the test submission, and I was trying to keep the impact as minimal as possible. I could probably do it tbf but I'm not really sure how having the logic moved into a reducer is really any different?

tomcrawleyy: i was thinking purely from a consistency standpoint that its going to make it harder to debug if theres logic handling tech record additions/deletions before submissions which isn't centralised

what refactor would you need to do to test submission, would it not just be dispatching an action in the same block as the code i referenced above? or am i missing something obvious which makes this more complicated

naathanbrown: Thanks Phil, will update and reformat this and come back with another pass :)

naathanbrown: Thanks lads, updated!

tomcrawleyy (approved):

pbardy2000 (approved):

PR #1408, opened 2024-02-26
feat(cb2-10920): Angular 16 to 17

pbardy2000:
## Upgrade angular 15 to 17

[CB2-10920](https://dvsa.atlassian.net/browse/CB2-10920)

## Checklist


tomcrawleyy (dismissed):

matthew2564 (dismissed):

src/app/features/tech-record/components/tech-record-title/tech-record-title.component.ts
```diff
@@ -18,7 +18,7 @@ import { Observable, take } from 'rxjs';
 })
 export class TechRecordTitleComponent implements OnInit {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  @Input() vehicle?: V3TechRecordModel | any;
+  @Input() vehicle?: any;
```
naathanbrown, line 21: how come we had to down grade this?

naathanbrown:
Happy with this, but I believe the bump to 16 introduced a bug. 
image
in the console

Quick look online led me to https://stackoverflow.com/questions/76429842/how-to-find-after-angular-upgrade-16-ng0600-writing-to-signals-is-not-allowed and appears to be a 16 introduction.

I've not really looked any further, but if it's easy reckon we could slip it in with this? Saves us writing a ticket etc. Again if it's hard ignore and go ticket path

src/app/features/tech-record/components/tech-record-title/tech-record-title.component.ts
pbardy2000: This isn't a downgrade, the new eslint package is much more powerful and basically says 'any subsumes the union' or something - basically if you or an any it doesn't matter, the type will always be any.

pbardy2000:
> Happy with this, but I believe the bump to 16 introduced a bug. image in the console
> 
> Quick look online led me to https://stackoverflow.com/questions/76429842/how-to-find-after-angular-upgrade-16-ng0600-writing-to-signals-is-not-allowed and appears to be a 16 introduction.
> 
> I've not really looked any further, but if it's easy reckon we could slip it in with this? Saves us writing a ticket etc. Again if it's hard ignore and go ticket path

It may have done, this one is confusing as, although Angular 16 introduced signals I don't think we're using them anywhere unless ngrx is using them under the hood. I assume this might be uncovering a different issue where we're probably do some bad practice regarding ngrx, but will investigate a bit more.

pbardy2000:
> Happy with this, but I believe the bump to 16 introduced a bug. image in the console
> 
> Quick look online led me to https://stackoverflow.com/questions/76429842/how-to-find-after-angular-upgrade-16-ng0600-writing-to-signals-is-not-allowed and appears to be a 16 introduction.
> 
> I've not really looked any further, but if it's easy reckon we could slip it in with this? Saves us writing a ticket etc. Again if it's hard ignore and go ticket path

Cheers for that, that page linked to this github issue https://github.com/ngrx/platform/issues/3892#issuecomment-1542791068, so I've just applied the fix they suggested which seems to have worked. I think the issue probably arises from the fact that we have a pipe doing a lot of data fetching and interacting with the store which is a bit strange to me but don't want mess around with it too much.

src/app/store/technical-records/reducers/technical-record-service.reducer.ts
```diff
@@ -480,11 +480,13 @@ function handleClearADRDetails(state: TechnicalRecordServiceState) {
 
       // If tank details 'statement' selected, null UN numbers, product list referene no., product list
       const { techRecord_adrDetails_tank_tankDetails_tankStatement_select: select } = sanitisedEditingTechRecord;
+      //  eslint-disable-next-line @typescript-eslint/no-unsafe-enum-comparison
```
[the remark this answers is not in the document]
pbardy2000: After discussion with some devs have turned off and added a comment to the eslint tech debt ticket to revisit at a later date when we can fix these things properly.

JunYanBJSS (approved):

PR #1420, opened 2024-03-01
DevOps - Create Manifest (52)

cvs-ops:
## Add Manifest.json file to the root of all CVS Functional Repositories


cvs-ops:
Abandoned by Workflow

PR #1454, opened 2024-04-09
fix(cb2-11068): fix tyre use code not showing all letters

pbardy2000:
## Tyre use code is not getting displayed in correctly

Set min-width to unset so select is not too large, but don't provide a max-width to allow the select to expand with text on different screen sizes, preventing cut-off.

[CB2-XXXX](https://dvsa.atlassian.net/browse/CB2-11068)

## Checklist

### DIFF
--- a/src/app/forms/components/select/select.component.scss
+++ b/src/app/forms/components/select/select.component.scss
@@ -30,6 +30,10 @@
   min-width: unset;
 }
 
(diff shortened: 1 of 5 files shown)


pbardy2000:
Closing as this will be merged into feature/RELEASE-123 instead.

PR #1459, opened 2024-04-09
fix(cb2-11185): break text to ensure correct text wrapping

pbardy2000:
## Bug - Max character limit on Tank Details Product List Reference Number shifts and breaks format of UN Codes

Setting the word-break to break-all ensures single character strings that are too long to fit on a single line are still broken and wrap, ensuring they fit in their parent containing without overflow or cutting-off.

[CB2-11185](https://dvsa.atlassian.net/browse/CB2-11185)

## Checklist


pbardy2000:
Closing as this will be merged into the feature branch.

PR #1577, opened 2024-09-09
fix(cb2-13837): handle fetching a specific trailer make which is miss…

pbardy2000:
…ing from reference data

## VTM Sentry - Cannot read properties of undefined (reading 'toString')

[CB2-13837](https://dvsa.atlassian.net/browse/CB2-13837)

## Checklist

### DIFF
--- a/src/app/store/reference-data/reference-data.selectors.ts
+++ b/src/app/store/reference-data/reference-data.selectors.ts
@@ -1,4 +1,9 @@
-import { Brake, ReferenceDataResourceType, ReferenceDataResourceTypeAudit } from '@models/reference-data.model';
+import {
+	Brake,
+	ReferenceDataModelBase,
+	ReferenceDataResourceType,
+	ReferenceDataResourceTypeAudit,
+} from '@models/reference-data.model';
 import { VehicleTypes } from '@models/vehicle-tech-record.model';
 import { createSelector } from '@ngrx/store';
 import {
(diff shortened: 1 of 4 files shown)


tomevs88 (dismissed):

src/app/store/reference-data/reference-data.selectors.ts
```diff
@@ -68,19 +73,25 @@ export const selectRefDataBySearchTerm = (
...
 				searchItem.push(obj);
 			}
 		});
+
 		if (searchTerm.length > 0) {
 			return searchItem;
 		}
+
 		return undefined;
 	});
```
matthew2564: I wonder if we look at using array methods instead of pushing into `searchItem` and then checking the length. At the same time, we might then look at introducing early returns, something like 
```
export const selectRefDataBySearchTerm = (
	searchTerm: string,
	referenceDataType: ReferenceDataResourceType,
	filter: keyof ReferenceDataModelBase
) =>
	createSelector(referenceDataFeatureState, (state): ReferenceDataModelBase[] => {
		const entities = state[referenceDataType]?.entities;

		if (!searchTerm || !entities) return [];

		return Object.values(entities).filter(
			(obj): obj is ReferenceDataModelBase =>
				obj !== undefined && (obj[filter]?.toString().toUpperCase() || '').includes(searchTerm.toUpperCase())
		);
	});
```
Maybe?

pbardy2000: Yeah that syntax looks nicer and removes the possibility of it returning undefined, so will apply that and the stricter typings for the areas that call it.

matthew2564 (approved):

BrandonT95 (approved):

PR #1629, opened 2024-11-01
fix(cb2-14944): show body declaration under specific conditions

pbardy2000:
## VTM - Selecting an ADR Body that is not apart of the approved list shows the Body Declaration field

[CB2-14944](https://dvsa.atlassian.net/browse/CB2-14944)

## Checklist

### DIFF
--- a/src/app/forms/validators/adr/adr.validators.ts
+++ b/src/app/forms/validators/adr/adr.validators.ts
@@ -84,12 +84,13 @@ export class AdrValidators {
 	 */
 	static setBodyDeclarationVisibility = (): ValidatorFn => {
 		return (control: AbstractControl) => {
-			const carriesDangerousGoods = control.parent?.get('techRecord_adrDetails_dangerousGoods')?.value;
-			const permittedDangerousGoods = control.parent?.get('techRecord_adrDetails_permittedDangerousGoods')?.value;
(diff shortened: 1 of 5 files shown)


src/app/forms/validators/adr/adr.validators.ts
```diff
@@ -67,4 +69,43 @@ export class AdrValidators {
...
+	 *    > Semi trailer box body
+	 *  - AND The permitted dangerous good 'Explosive (type 3)' is selected
+	 *
+	 * @param control
+	 */
+	static setBodyDeclarationVisibility = (): ValidatorFn => {
+		return (control: AbstractControl) => {
+			const carriesDangerousGoods = control.parent?.get('techRecord_adrDetails_dangerousGoods')?.value;
+			const permittedDangerousGoods = control.parent?.get('techRecord_adrDetails_permittedDangerousGoods')?.value;
+			const adrBodyType = control.parent?.get('techRecord_adrDetails_vehicleDetails_type')?.value;
```
matthew2564: We could move these inside the `if` check, it seems they're only relevant when that check is satisfied

pbardy2000: Can do, suppose the form control name is already pretty descriptive so no need for the duplication.

matthew2564 (approved):

tomcrawleyy (approved):

PR #1958, opened 2025-07-18
feat(CB2-18355): GovUK Footer

tomcrawleyy:
## Ticket title

_One line description_
[CB2-18355](https://dvsa.atlassian.net/browse/CB2-18355)

## Checklist

[CB2-18355]: https://dvsa.atlassian.net/browse/CB2-18355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/core/components/footer/footer.component.html
```diff
@@ -1,4 +1,17 @@
-<footer class="govuk-footer" role="contentinfo">
+ <footer role="contentinfo">
+  <svg class="small-crown" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 64 64" width="64" height="64">
```
pbardy2000:
```suggestion
  <svg class="small-crown" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 64 64" width="32" height="30">
```

This should be inside the govuk-width-container so it looks correct on larger viewports.

src/app/core/components/footer/footer.component.scss
```diff
@@ -1,4 +1,13 @@
 :host {
+  footer {
+    border-top: 10px #1d70b8 solid;
+    padding-top: 30px;
+  }
+
+  .small-crown {
+    padding-left: 30px;
+  }
```
pbardy2000: This isn't need if the svg is places inside the govuk-width-container as it will then wrap correctly, you can use gov.uk as a reference.

src/app/core/components/footer/footer.component.html
```diff
@@ -1,4 +1,17 @@
-<footer class="govuk-footer" role="contentinfo">
+ <footer role="contentinfo">
```
pbardy2000:
```suggestion
 <footer class="govuk-footer" role="contentinfo">
```

Removing this class breaks the font-family of the bottom-most text.

pbardy2000 (approved):

BrandonT95 (approved):

PR #2173, opened 2025-11-21
feat(CB2-19847): add filter functionality

pbardy2000:
## VTM - Redesign - HGV 'Plates' filters (first ticket) [FF]

[CB2-19847](https://dvsa.atlassian.net/browse/CB2-19847)

## Checklist

[CB2-19847]: https://dvsa.atlassian.net/browse/CB2-19847?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/directives/filter-by-tags/filter-by-tags.directive.ts
+++ b/src/app/directives/filter-by-tags/filter-by-tags.directive.ts
@@ -19,16 +19,16 @@ export class FilterByTagsDirective {
 
 			// If no filters are applied, show the element
 			if (filters.length === 0) {
-				return this.renderer.setStyle(this.elementRef.nativeElement, 'display', 'initial');
+				return this.renderer.setStyle(element, 'display', 'initial');
 			}
 
 			// If filters have been applied, but none match the tags, hide the element
-			if (filters.some((filter) => !tagNames.includes(filter))) {
-				return this.renderer.setStyle(this.elementRef.nativeElement, 'display', 'none');
+			if (filters.every((filter) => !tagNames.includes(filter))) {
+				return this.renderer.setStyle(element, 'display', 'none');
 			}
 
 			// Otherwise, show the element
-			this.renderer.setStyle(this.elementRef.nativeElement, 'display', 'initial');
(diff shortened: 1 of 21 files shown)


tomevs88 (approved):

src/app/directives/filter-by-tags/filter-by-tags.directive.ts
```diff
@@ -0,0 +1,34 @@
...
+
+			// If filters have been applied, but none match the tags, hide the element
+			if (filters.every((filter) => !tagNames.includes(filter))) {
+				return this.renderer.setStyle(element, 'display', 'none');
+			}
+
+			// Otherwise, show the element
+			this.renderer.setStyle(element, 'display', 'initial');
+		});
+	}
```
matthew2564, line 35: Is there any benefit to resolving the reactive values on lines 15/16 when the early returns are in play? If not, would something like this work

```
		effect(() => {
			const element = this.elementRef.nativeElement;
			if (!element) return;

	      	const filters = this.filters();

			// If no filters are applied, show the element
			if (filters.length === 0) {
				return this.renderer.setStyle(element, 'display', 'initial');
			}

      		const tagNames = this.tagNames();

			// If filters have been applied, but none match the tags, hide the element
			if (filters.every((filter) => !tagNames.includes(filter))) {
				return this.renderer.setStyle(element, 'display', 'none');
			}

			// Otherwise, show the element
			this.renderer.setStyle(element, 'display', 'initial');
		});
```

Would we benefit from an early return with tags when empty too, do we know how many filters we're likely to have?

matthew2564 (approved):

src/app/directives/filter-by-tags/filter-by-tags.directive.ts
pbardy2000: Not really, the effect will know to rerun whenever the dependent signals update regardless of where they are in the callback. Have updated to your suggestion as it works the same way.

